PR #4, opened 2016-04-12
Create Shared Library on all platforms

bslassey:
### Pull Request check-list

_Please make sure to review and check all of these items:_
- [ ] Does `make -j8 test` (UNIX) or `vcbuild test nosign` (Windows) pass with
  this change (including linting)?
- [ ] Is the commit message formatted according to [CONTRIBUTING.md](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#step-3-commit)?
- [ ] If this change fixes a bug (or a performance problem), is a regression
  test (or a benchmark) included?
- [ ] Is a documentation update included (if this change modifies
  existing APIs, or introduces new ones)?

_NOTE: these things are not required to open a PR and can be done
afterwards / while the PR is open._
### Affected core subsystem(s)

_Please provide affected core subsystem(s) (like buffer, cluster, crypto, etc)_
### Description of change

_Please provide a description of the change here._

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -38,7 +38,7 @@
         'V8_BASE': '<(PRODUCT_DIR)/libv8_base.a',
       }, {
         'OBJ_DIR': '<(PRODUCT_DIR)/obj.target',
-        'V8_BASE': '<(PRODUCT_DIR)/obj.target/deps/v8/tools/gyp/libv8_base.a',
+        'V8_BASE': '<(PRODUCT_DIR)/obj.target/deps/spidershim/libspidershim.a',
       }],
       ['openssl_fips != ""', {
         'OPENSSL_PRODUCT': 'libcrypto.a',
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -104,6 +104,8 @@ template <class T, class M = NonCopyablePersistentTraits<T> > class Persistent;
 template <typename T> class FunctionCallbackInfo;
 template <typename T> class PropertyCallbackInfo;
 
+ enum JsErrorCode : unsigned int;
+ 
 class JitCodeEvent;
 class RetainedObjectInfo;
 struct ExternalArrayData;
@@ -478,13 +480,13 @@ struct BarrierMethods {
 template <class T>
 class PersistentBase {
  public:
-  V8_INLINE void Reset();
+  void Reset();
 
   template <class S>
-  V8_INLINE void Reset(Isolate* isolate, const Handle<S>& other);
+  void Reset(Isolate* isolate, const Handle<S>& other);
 
   template <class S>
-  V8_INLINE void Reset(Isolate* isolate, const PersistentBase<S>& other);
+  void Reset(Isolate* isolate, const PersistentBase<S>& other);
 
   V8_INLINE bool IsEmpty() const { return val_ == NULL; }
   V8_INLINE void Empty() { Reset(); }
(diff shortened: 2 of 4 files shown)


common.gypi
```diff
@@ -38,7 +38,7 @@
         'V8_BASE': '<(PRODUCT_DIR)/libv8_base.a',
       }, {
         'OBJ_DIR': '<(PRODUCT_DIR)/obj.target',
-        'V8_BASE': '<(PRODUCT_DIR)/obj.target/deps/v8/tools/gyp/libv8_base.a',
+        'V8_BASE': '<(PRODUCT_DIR)/obj.target/deps/spidershim/libspidershim.a',
```
ehsan, line 41: This is not quite right, since it breaks v8 based builds.  I think this commit is probably obsolete with 518ddfc1.

deps/spidershim/include/v8.h
```diff
@@ -104,6 +104,8 @@ template <class T, class M = NonCopyablePersistentTraits<T> > class Persistent;
 template <typename T> class FunctionCallbackInfo;
 template <typename T> class PropertyCallbackInfo;
 
+ enum JsErrorCode : unsigned int;
```
ehsan, line 107: I fixed this in e129b77e.

deps/spidershim/include/v8.h
```diff
@@ -478,13 +480,13 @@ struct BarrierMethods {
 template <class T>
 class PersistentBase {
  public:
-  V8_INLINE void Reset();
+  void Reset();
```
ehsan, line 483: Why are you removing V8_INLINE's here?  There's some functions that are inlined at the end of this file, and these ones can be inlined there too.

deps/spidershim/src/v8conversions.h
```diff
@@ -21,6 +21,7 @@
 #pragma once
 #include <stdint.h>
 #include <stdlib.h>
+#include <cstring>
```
ehsan, line 24: This part looks good!  Committed in d30d4f8.

deps/spidershim/spidermonkey/mozglue/build/moz.build
```diff
@@ -6,10 +6,7 @@
 
 # Build mozglue as a shared lib on Windows, OSX and Android.
 # If this is ever changed, update MOZ_SHARED_MOZGLUE in browser/installer/Makefile.in
-if CONFIG['OS_TARGET'] in ('WINNT', 'Darwin', 'Android'):
-    SharedLibrary('mozglue')
-else:
-    Library('mozglue')
+SharedLibrary('mozglue')
```
ehsan, line 9: Hmm, why do you need to do this?  I wonder if there's a better way to fix the issue you're facing?

ehsan:
Thanks for the PR!  I submitted a bunch of comments, can you please answer/address them?  Thanks!

deps/spidershim/spidermonkey/mozglue/build/moz.build
bslassey, line 9: The build expects libmozglue.so to be in deps/spidershim/build/dist/bin, but on Linux because we have Library('mozglue') we produce libmozglue.a.desc instead.

This is why the build was working on your Mac but not for people building on Linux.

deps/spidershim/include/v8.h
bslassey, line 483: The compiler was throwing errors on inlining functions without bodies. This might be a mac/linux compiler difference.

common.gypi
ehsan, line 41: This should be fixed correctly in 21ff77a2.

deps/spidershim/spidermonkey/mozglue/build/moz.build
ehsan, line 9: I see, makes sense.  But the right fix is to link against mozglue statically on Linux.  Fixed that in 2df4ce2225.

deps/spidershim/include/v8.h
ehsan, line 483: OK, I reproduced and fixed them in 7a3002c.  Thanks!

ehsan:
Try building on Linux now from master.  For me, in both debug and non-debug mode, I can build until linking the node binary, just the same as Mac now.

Closing this PR since I think everything it used to do is fixed on master now.

PR #6, opened 2016-04-15
set debug flags based on build type (configuration) (fixes #5)

### LINKED ISSUE #5: Figure out how to support building spidermonkey in debug mode
I tried to fight gyp in order to support this to no avail.  Here is what I tried:

```
  'target_defaults': {
    'configurations': {
      'Debug': {
        'variables': { 'spidermonkey_build_type': 'debug' },
      },
      ...
    },
  },
```

But it seems like variables defined in this context cannot be used outside of the target_defaults, which means I won't be able to add a `<(spidermonkey_build_type)>` to the action, etc.

mykmelez:
Gyp sets $BUILDTYPE to the build configuration name (Release, Debug), and this sets SpiderMonkey's debug flags (--enable-debug, --disable-optimize) based on that variable. So it should build with these flags when you configure Gyp to also build debug via `./configure --enable-engine=spidermonkey --debug`. But I haven't yet gotten Gyp to actually do a debug build yet, so I can't confirm that it works.

### DIFF
--- a/deps/spidershim/build-spidermonkey.sh
+++ b/deps/spidershim/build-spidermonkey.sh
@@ -13,8 +13,12 @@ if test -z "$AUTOCONF"; then
   exit 1
 fi
 
+if test "$BUILDTYPE" = "Debug"; then
+  DEBUG_FLAGS="--enable-debug --disable-optimize"
+fi
+
 test -d build || mkdir build
 cd build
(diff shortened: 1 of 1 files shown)


mykmelez:
Closing, as @ehsan fixed #5 a different way in 8ccb61cddc500bfd85af60b272ae8d163dec5911.

PR #9, opened 2016-04-16
Implement v8global

evilpie:
Implements v8::{Undefined, Null, True, False} and stubs SetResourceConstraints like Chakra.


ehsan:
Thanks! Are you also planing to write a test?  You can add it to https://github.com/mozilla/spidernode/blob/master/deps/spidershim/test/value.cc for example.

(Note that in order to build the tests, you can use `make -k` to build as much of the dependency tree as possible, and then run `./out/{Debug,Release}/testname` manually.  Current tests are `hello-world` and `value`.)

PR #37, opened 2016-04-21
ignore the build-type-specific dirs

mykmelez:
Builds build in build-opt or build-debug now, so this updates the SpiderShim-specific .gitignore file to ignore those dirs.


ehsan:
Thanks!

PR #39, opened 2016-04-22
spidernode: implement ArrayBuffer::New(Isolate*, size_t)

tbsaunde:

### DIFF
--- a/deps/spidershim/src/v8arraybuffer.cc
+++ b/deps/spidershim/src/v8arraybuffer.cc
@@ -1,0 +1,40 @@
+// Copyright Mozilla Foundation. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


deps/spidershim/src/v8arraybuffer.cc
```diff
@@ -0,0 +1,47 @@
...
+
+#include "v8.h"
+#include "v8conversions.h"
+#include "v8local.h"
+#include "jsapi.h"
+#include "jsfriendapi.h"
+#include "js/Conversions.h"
+
+static_assert(sizeof(v8::Value) == sizeof(JS::Value),
+              "v8::Value and JS::Value must be binary compatible");
```
ehsan: We have this assertion elsewhere, it's not needed here, right?

deps/spidershim/src/v8arraybuffer.cc
```diff
@@ -0,0 +1,47 @@
...
+
+namespace v8 {
+
+Local<ArrayBuffer>
+ArrayBuffer::New(Isolate* isolate, size_t size)
+{
+	if (!isolate) {
```
ehsan: OMG tabs :((

I'll need to clang-format immediately after merging your commit, I guess!

deps/spidershim/src/v8arraybuffer.cc
```diff
@@ -0,0 +1,47 @@
...
+		isolate = Isolate::GetCurrent();
+	}
```
ehsan: Null isolate's are not needed here, so you don't need this if block at all.

ehsan:
Can you also resolve the merge conflicts while you're addressing the comments?  Thanks!

ehsan:
Thanks!

PR #40, opened 2016-04-22
implement String::NewFromOneByte and String::NewFromTwoByte

mykmelez:
This PR implements _String::NewFromOneByte_ and _String::NewFromTwoByte_. It also adds a test that demonstrates a bug in _String::NewFromUtf8_, which currently uses _JS_NewStringCopyN_ and _JS_NewStringCopyZ_ to decode the input data, although those functions interpret the data as Latin-1, not UTF-8. (I'm not yet sure what the fix is, but I figured it'd still be worth adding a failing test for the bug, so we can confirm the fix once we land it.)

### DIFF
--- a/deps/spidershim/test/value.cc
+++ b/deps/spidershim/test/value.cc
@@ -828,16 +828,6 @@ TEST(SpiderShim, String) {
   const uint16_t utf16Data[] = { 0x02E4, 0x0064, 0x12E4, 0x0030, 0x3045, 0x0000 };
   const unsigned char utf8Data[] = { 0xCB, 0xA4, 0x64, 0xE1, 0x8B, 0xA4, 0x30, 0xE3, 0x81, 0x85, 0x00 };
 
-  // This test currently fails because String::NewFromUtf8 interprets its input
-  // as Latin-1 rather than UTF-8.
-  Local<String> fromUtf8Str =
-    String::NewFromUtf8(engine.isolate(), reinterpret_cast<const char*>(utf8Data), NewStringType::kNormal).
-      ToLocalChecked();
-  EXPECT_EQ(5, fromUtf8Str->Length());
-  EXPECT_EQ(10, fromUtf8Str->Utf8Length());
-  String::Value fromUtf8Val(fromUtf8Str);
-  EXPECT_EQ(0, memcmp(*fromUtf8Val, utf16Data, sizeof(*utf16Data)));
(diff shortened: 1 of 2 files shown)


deps/spidershim/test/value.cc
```diff
@@ -795,6 +795,56 @@ TEST(SpiderShim, String) {
...
+  // U+02E4 -> CB A4
+  // U+0064 -> 64
+  // U+12E4 -> E1 8B A4
+  // U+0030 -> 30
+  // U+3045 -> E3 81 85
+  const uint16_t utf16Data[] = { 0x02E4, 0x0064, 0x12E4, 0x0030, 0x3045, 0x0000 };
+  const unsigned char utf8Data[] = { 0xCB, 0xA4, 0x64, 0xE1, 0x8B, 0xA4, 0x30, 0xE3, 0x81, 0x85, 0x00 };
+
+  // This test currently fails because String::NewFromUtf8 interprets its input
+  // as Latin-1 rather than UTF-8.
```
ehsan: Thanks for catching this!

But we can't commit a test that will make our Travis-CI builds all red.  Fixing this should be easy, I think, by using something like `JS::UTF8CharsToNewTwoByteCharsZ()` from `js/CharacterEncoding.h` first, and then using `JS_NewUCString`.  I think in that case we can just transfer the ownership of the buffer that `JS::UTF8CharsToNewTwoByteCharsZ()` to the JS engine, so we wouldn't need to copy the buffer twice.

Can you please either try to fix that bug, or take this test out of this commit and file a new issue for fixing the bug and adding this test with it?  (You can `#if 0` it out if you don't want to fix the bug here.)

ehsan:
Thanks @mykmelez, this looks great!  I'll merge once the test issue has been addressed.

deps/spidershim/test/value.cc
mykmelez: > But we can't commit a test that will make our Travis-CI builds all red.

Understood. I hadn't checked Travis, but another test was failing for me locally, so I thought perhaps we weren't yet counting on tests to pass in CI. I've filed #43 on that local failure and #44 to fix the String::NewFromUtf8 bug.

mykmelez:
> Thanks @mykmelez, this looks great! I'll merge once the test issue has been addressed.

I removed the failing test in 0689ecc.

PR #45, opened 2016-04-25
ensure String::NewFromUtf8 encodes as UTF-8 (fixes #44)

### LINKED ISSUE #44: String::NewFromUtf8 encodes to Latin-1 code points
String::NewFromUtf8 currently uses JS_NewStringCopyN and JS_NewStringCopyZ to encode its input data, but those functions interpret the data as Latin-1, not UTF-8, so it encodes the data as a sequence of Latin-1 code points instead of UTF-8 code points.

Per @ehsan in https://github.com/mozilla/spidernode/pull/40#discussion_r60796855, we should be able to use "something like JS::UTF8CharsToNewTwoByteCharsZ() from js/CharacterEncoding.h first, and then using JS_NewUCString. I think in that case we can just transfer the ownership of the buffer that JS::UTF8CharsToNewTwoByteCharsZ() to the JS engine, so we wouldn't need to copy the buffer twice."

mykmelez:
This fixes issue #44 by encoding input data as UTF-8 in String::NewFromUtf8.

### DIFF
--- a/deps/spidershim/src/v8string.cc
+++ b/deps/spidershim/src/v8string.cc
@@ -79,9 +79,20 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
                                        v8::NewStringType type, int length) {
   assert(type == v8::NewStringType::kNormal); // TODO: Add support for interned strings
   JSContext* cx = JSContextFromIsolate(isolate);
-  JS::RootedString str(cx, length >= 0 ?
-                        JS_NewStringCopyN(cx, data, length) :
-                        JS_NewStringCopyZ(cx, data));
+
+  // In V8's api.cc, this conditional block is annotated with the comment:
+  //    TODO(dcarney): throw a context free exception.
+  if (length > v8::String::kMaxLength) {
+    return MaybeLocal<String>();
+  }
+
+  if (length < 0) {
+    length = strlen(data);
+  }
+
+  size_t twoByteLen;
+  char16_t* twoByteChars = JS::UTF8CharsToNewTwoByteCharsZ(cx, JS::UTF8Chars(data, length), &twoByteLen).get();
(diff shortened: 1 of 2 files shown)


ehsan:
Thanks, LGTM!

deps/spidershim/src/v8string.cc
```diff
@@ -79,9 +79,20 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
...
+  if (length > v8::String::kMaxLength) {
+    return MaybeLocal<String>();
+  }
+
+  if (length < 0) {
+    length = strlen(data);
+  }
+
+  size_t twoByteLen;
+  char16_t* twoByteChars = JS::UTF8CharsToNewTwoByteCharsZ(cx, JS::UTF8Chars(data, length), &twoByteLen).get();
```
evilpie, line 94: Looks to to me like this is leaking. Could declare twoByteChars as UniqueTwoByteChars.

mykmelez, line 94: Hmm, I might misunderstand, but I think this only leaks when the JS_NewUCString call below fails, since that function takes ownership of twoByteChars on success, per https://dxr.mozilla.org/mozilla-central/rev/fc15477ce628599519cb0055f52cc195d640dc94/js/src/jsapi.h#4520-4524.

If this declared twoByteChars as UniqueTwoByteChars, then it would have to use JS_NewUCStringCopyN to copy twoByteChars instead of reusing it, as on this branch:

https://github.com/mozilla/spidernode/compare/master...mykmelez:free-two-byte-chars-automatically

And that seems like an unnecessary extra allocation. So I would rather explicitly free twoByteChars on error, as in pull request #47.

How does that sound?

ehsan, line 94: I responded to this [here](https://github.com/mozilla/spidernode/pull/47#issuecomment-214930086).

PR #47, opened 2016-04-26
free twoByteChars on JS_NewUCString error

mykmelez:
Per https://dxr.mozilla.org/mozilla-central/rev/fc15477ce628599519cb0055f52cc195d640dc94/js/src/jsapi.h#4520-4524. This also returns early on error in JS::UTF8CharsToNewTwoByteCharsZ, which can also fail.


mykmelez:
I noticed this when @evilpie suggested that twoByteChars is leaking over in https://github.com/mozilla/spidernode/pull/45/files/a185c810a9e124e6e805afe93d3fea5bea579451#r61071370. I might misunderstand, but I think it's only leaking on JS_NewUCString failure, hence this pull request. However, we could use UniqueTwoByteChars to free it, provided we switch to JS_NewUCStringCopyN, per this other branch:

https://github.com/mozilla/spidernode/compare/master...mykmelez:free-two-byte-chars-automatically

Let me know if you think we should do that instead.

ehsan:
@evilpie is right, we should be using `UniquePtr` (in this case `UniqueTwoByteChars`) everywhere.  The problem with this patch is that the next time someone adds another early bailout, the leak can be reintroduced.

UniquePtr supports a `release()` method that releases the ownership of the object back to the caller, so you can do something like `JS_NewUCString(..., ptr.release(), ...)` to transfer the ownership.  Do you mind giving that a shot, please?  Thanks!

mykmelez:
> UniquePtr supports a release() method that releases the ownership of the object back to the caller, so you can do something like JS_NewUCString(..., ptr.release(), ...) to transfer the ownership. Do you mind giving that a shot, please? Thanks!

Got it, makes sense. I've done that in 8bbdb21.

ehsan:
Thanks!

PR #48, opened 2016-04-26
make expected value be the first argument to EXPECT_EQ/STREQ

mykmelez:
EXPECT_EQ/EXPECT_STREQ expect their first argument to be the expected value.

https://github.com/mozilla/spidernode/blob/e0ca6e7379230698bf0434b8028a98caabe06a8c/deps/gtest/include/gtest/gtest.h#L2015
https://github.com/mozilla/spidernode/blob/e0ca6e7379230698bf0434b8028a98caabe06a8c/deps/gtest/include/gtest/gtest.h#L2088

So they output strange error messages when the first value of a failing assertion is the actual value. For example, given this code:

```
TEST(SpiderShim, Test) {
  int actual_value = 10;

  // Expected before actual.
  EXPECT_EQ(5, actual_value);

  // Actual before expected.
  EXPECT_EQ(actual_value, 5);
}
```

The first failure outputs:

> ../deps/spidershim/test/value.cc:18: Failure
> Value of: actual_value
>  Actual: 10
> Expected: 5

While the second one outputs:

> ../deps/spidershim/test/value.cc:21: Failure
> Value of: 5
> Expected: actual_value
> Which is: 10

So value.cc should pass the expected value as the first argument to those macros. It already does, in a few cases. This branch fixes the rest.

### DIFF
--- a/deps/spidershim/test/value.cc
+++ b/deps/spidershim/test/value.cc
@@ -16,23 +16,23 @@ void TestBoolean(Isolate* isolate, bool value) {
   EXPECT_TRUE(boolean->IsBoolean());
   EXPECT_FALSE(boolean->IsNumber());
   EXPECT_FALSE(boolean->IsBooleanObject());
-  EXPECT_EQ(boolean->IsTrue(), value);
-  EXPECT_EQ(boolean->IsFalse(), !value);
-  EXPECT_EQ(boolean->Value(), value);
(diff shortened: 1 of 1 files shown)


ehsan:
Thanks so much!  I was wondering why I'm getting nonsensical error messages sometimes.  My brain has been corrupted by the SimpleTest.js APIs.  :-)

PR #49, opened 2016-04-28
implement String::NewExternalTwoByte

mykmelez:
This branch implements String::NewExternalTwoByte by translating between the V8 API, which represents an external string as an ExternalStringResource instance whose destructor frees its resources; and the SpiderMonkey API, which registers a JSStringFinalizer struct for the string whose finalize() method does that work.

All examples of JSStringFinalizer in Gecko are static, but this implementation needs to associate them with arbitrary ExternalStringResource instances, so it extends JSStringFinalizer to have a resource_ member and instantiate the struct dynamically. Presumably this is the use case for which [bug 776000](https://bugzilla.mozilla.org/show_bug.cgi?id=776000) was filed, and this implementation may be somewhat akin to [comment 8 on that bug](https://bugzilla.mozilla.org/show_bug.cgi?id=776000#c8).

I'd love to get design input in additional to code review, as it isn't clear to me if this is the best possible way to translate between the two APIs.

Note: the v8.h changes simply revert changes made in deps/chakrashim/include/v8.h, so ExternalStringResource and ExternalStringResourceBase are defined the way they are in deps/v8/include/v8.h. [ChakraShim's version of String::NewExternalTwoByte](https://github.com/mozilla/spidernode/blob/e0ca6e7379230698bf0434b8028a98caabe06a8c/deps/chakrashim/src/v8string.cc#L315-L328) doesn't currently manage external strings, it simply copies their data and deletes them. Which is presumably why [this ChakraShim comment](https://github.com/mozilla/spidernode/blob/e0ca6e7379230698bf0434b8028a98caabe06a8c/src/string_bytes.cc#L79-L81) warns against accessing a string after passing it to String::NewExternal.

Also, the test could be more thorough, although I'm unsure how to expand it. I haven't found much existing code for testing external strings (although they're used incidentally to test other APIs).

### DIFF
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -113,6 +113,7 @@ struct ExternalArrayData;
 namespace internal {
 class RootStore;
 template <class T> class Local;
+struct ExternalStringFinalizer;
 }
 
 enum PropertyAttribute {
@@ -1171,7 +1172,7 @@ class V8_EXPORT String : public Name {
     ExternalStringResourceBase(const ExternalStringResourceBase&);
     void operator=(const ExternalStringResourceBase&);
 
-    friend struct ExternalStringFinalizer;
+    friend struct internal::ExternalStringFinalizer;
   };
 
   /**
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -106,9 +106,6 @@ template <class T, class M = NonCopyablePersistentTraits<T> > class Persistent;
 template <typename T> class FunctionCallbackInfo;
 template <typename T> class PropertyCallbackInfo;
 
-namespace internal {
-class Heap;
-}  // namespace internal
 class JitCodeEvent;
 class RetainedObjectInfo;
 struct ExternalArrayData;
@@ -1174,7 +1171,6 @@ class V8_EXPORT String : public Name {
     ExternalStringResourceBase(const ExternalStringResourceBase&);
     void operator=(const ExternalStringResourceBase&);
 
-    friend class v8::internal::Heap;
     friend struct ExternalStringFinalizer;
   };
 
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -106,6 +106,9 @@ template <class T, class M = NonCopyablePersistentTraits<T> > class Persistent;
 template <typename T> class FunctionCallbackInfo;
 template <typename T> class PropertyCallbackInfo;
 
+namespace internal {
+class Heap;
+}  // namespace internal
 class JitCodeEvent;
 class RetainedObjectInfo;
 struct ExternalArrayData;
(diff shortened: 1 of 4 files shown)


ehsan:
Sorry I didn't get a chance to look at this.  Will take a look tomorrow!

deps/spidershim/include/v8.h
```diff
@@ -1149,18 +1152,85 @@ class V8_EXPORT String : public Name {
...
+     * control how allocated external string resources are disposed.
+     */
+    virtual void Dispose() { delete this; }
+
+   private:
+    // Disallow copying and assigning.
+    ExternalStringResourceBase(const ExternalStringResourceBase&);
+    void operator=(const ExternalStringResourceBase&);
+
+    friend class v8::internal::Heap;
```
ehsan: Since we have our own `v8::internal` namespace, please delete this friend, and also the forward declaration of the `Heap` class in the beginning of `v8.h`.

deps/spidershim/src/v8string.cc
```diff
@@ -137,6 +137,60 @@ Local<String> String::NewFromTwoByte(Isolate* isolate, const uint16_t* data,
                         length).FromMaybe(Local<String>());
 }
 
+struct ExternalStringFinalizer : JSStringFinalizer {
```
ehsan: Nit: I've been trying to avoid adding non-public symbols to the `v8` namespace.  Please move this to either `v8::internal` or an anonymous namespace.

deps/spidershim/src/v8string.cc
```diff
@@ -137,6 +137,60 @@ Local<String> String::NewFromTwoByte(Isolate* isolate, const uint16_t* data,
...
+      this->resource_ = nullptr;
+    }
+
+    // Delete ourselves, which JSExternalString::finalize doesn't do,
+    // presumably because it assumes we're static.
+    delete this;
+  };
+};
+
+void FinalizeExternalString(const JSStringFinalizer* fin, char16_t* chars) {
```
ehsan: Since this conceptually belongs to `ExternalStringFinalizer`, can you please make it a static method of that class?

deps/spidershim/src/v8string.cc
```diff
@@ -137,6 +137,60 @@ Local<String> String::NewFromTwoByte(Isolate* isolate, const uint16_t* data,
...
+MaybeLocal<String> String::NewExternalTwoByte(Isolate* isolate,
+                                              ExternalStringResource* resource) {
+  JSContext* cx = JSContextFromIsolate(isolate);
+
+  auto fin = mozilla::MakeUnique<ExternalStringFinalizer>(resource);
+  if (!fin) {
+    return MaybeLocal<String>();
+  }
+
+  fin->finalize = FinalizeExternalString;
```
ehsan: It seems like this can be moved to the constructor of `ExternalStringFinalizer`.

deps/spidershim/test/value.cc
```diff
@@ -778,6 +778,24 @@ TEST(SpiderShim, Date) {
   EXPECT_EQ(time, Date::Cast(*date.ToLocalChecked()->ToObject())->ValueOf());
 }
 
+namespace {
```
ehsan, line 781: There are a number of tests about external strings in v8's `test-api.cc`, such as `ScriptUsingStringResource`, `ScriptUsingOneByteStringResource`, and `ScriptMakingExternalString`.  Out of curiosity, have you tried adopting those tests?  Massaging the tests a bit to get them to compile is fine.  And if they have parts that depend on features we don't support yet, you can comment them out too.

(FWIW this can happen in a different PR, not going to block this on porting those tests, but I've been trying to adopt v8 tests where possible, to make sure we agree on the API semantics and whatnot.)

deps/spidershim/test/value.cc
```diff
@@ -778,6 +778,24 @@ TEST(SpiderShim, Date) {
...
+
+class TestExternalStringResource : public String::ExternalStringResource {
+  public:
+    TestExternalStringResource(const uint16_t* source, size_t length)
+        : data_(source), length_(length) {}
+    ~TestExternalStringResource() {
+      // XXX Check that we get called when the string is finalized.
```
ehsan: The v8 tests have a trick to test this by passing a reference to a counter and store it in the object, and increment the counter in the destructor, and forcing a GC and checking the counter value.  We don't have anything exposed to force a SpiderMonkey GC yet and I prefer to keep it that way if we can.  However, you can pass in a pointer to a global `bool` or something like that, and add an `atexit` callback that checks whether that `bool` has been set as a result of this finalizer being called.  That would be some dark magic but I think it should work reliably.

ehsan:
Thanks a lot for this, @mykmelez.  I think you have come up with a great solution for this, and I can't think of anything better we could do here without [bug 776000](https://bugzilla.mozilla.org/show_bug.cgi?id=776000), so the approach is fine.  I have some comment below on the code and the test but they're all relatively minor!  I'll merge after you have addressed these.

deps/spidershim/include/v8.h
mykmelez: > Since we have our own v8::internal namespace, please delete this friend, and also the forward declaration of the Heap class in the beginning of v8.h.

Good point, done in 68e6305.

deps/spidershim/src/v8string.cc
mykmelez: > Nit: I've been trying to avoid adding non-public symbols to the v8 namespace. Please move this to either v8::internal or an anonymous namespace.

Makes sense, I moved it to v8::internal in 5bcd5eb.

deps/spidershim/src/v8string.cc
mykmelez: > Since this conceptually belongs to ExternalStringFinalizer, can you please make it a static method of that class?

Indeed, 3f68a5a does so.

deps/spidershim/src/v8string.cc
mykmelez: > It seems like this can be moved to the constructor of ExternalStringFinalizer.

Yes, and that's a better encapsulation of the initialization of the struct. I moved it in c6257b6.

deps/spidershim/test/value.cc
mykmelez, line 781: > There are a number of tests about external strings in v8's test-api.cc, such as ScriptUsingStringResource, ScriptUsingOneByteStringResource, and ScriptMakingExternalString. Out of curiosity, have you tried adopting those tests? Massaging the tests a bit to get them to compile is fine. And if they have parts that depend on features we don't support yet, you can comment them out too.

I looked at those, but it wasn't clear at first glance how to port them. I'll take another look, however.

> (FWIW this can happen in a different PR, not going to block this on porting those tests, but I've been trying to adopt v8 tests where possible, to make sure we agree on the API semantics and whatnot.)

I totally agree! The ideal end state would be to be able to run all of V8 tests verbatim, although that's a long-term goal, for which porting the tests is a good first step.

deps/spidershim/test/value.cc
mykmelez: > you can pass in a pointer to a global bool or something like that, and add an atexit callback that checks whether that bool has been set as a result of this finalizer being called. That would be some dark magic but I think it should work reliably.

That works in theory, except that I don't see a way to hook an atexit check into the gtest harness. Of course I could simply _printf_ the result, but if I'm only going to do that, then I might as well simply _printf_ inside the destructor itself (which is what I do when manually testing this branch).

Perhaps instead I can define a TEST(SpiderShim, ExternalStringDestructor) function that checks the result of the destructor. The test harness appears to run the tests in deterministic order (and reconstruct the engine, or at least the isolate, for each test function), so that should work.

mykmelez: > Perhaps instead I can define a TEST(SpiderShim, ExternalStringDestructor) function that checks the result of the destructor. The test harness appears to run the tests in deterministic order (and reconstruct the engine, or at least the isolate, for each test function), so that should work.

Indeed, it does seem to work, per 7f07aef. Is it a reasonable approach?

deps/spidershim/test/value.cc
mykmelez, line 781: > I looked at those, but it wasn't clear at first glance how to port them. I'll take another look, however.

To close the loop on this: I took another look, and it's more clear how to port those tests now, but it isn't clear how useful it is to do so, given that they mostly test features that we haven't implemented yet (and that Node doesn't use). Nevertheless, I've started porting them, with the first two in #59.

PR #53, opened 2016-05-04
implement String::Write*

mykmelez:
This branch implements the String::Write, String::WriteOneByte, and String:WriteUtf8 methods, the last three unimplemented V8 String::\* methods that Node uses. It includes unit tests cribbed from V8's test suite, to confirm that its behavior conforms to V8's expectations.

### DIFF
--- a/deps/spidershim/test/value.cc
+++ b/deps/spidershim/test/value.cc
@@ -11,6 +11,32 @@
 
 #include "gtest/gtest.h"
 
+// From test-api.cc.
+static int StrCmp16(uint16_t* a, uint16_t* b) {
+  while (true) {
+    if (*a == 0 && *b == 0) return 0;
+    if (*a != *b) return 0 + *a - *b;
+    a++;
+    b++;
+  }
+}
+
+// From test-api.cc.
+static int StrNCmp16(uint16_t* a, uint16_t* b, int n) {
+  while (true) {
+    if (n-- == 0) return 0;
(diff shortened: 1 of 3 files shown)


deps/spidershim/test/value.cc
```diff
@@ -921,6 +947,361 @@ TEST(SpiderShim, ExternalStringResourceDestructorCalled) {
   EXPECT_TRUE(externalOneByteStringResourceDestructorCalled);
 }
 
+// Begin code copied from THREADED_TEST(StringWrite) in test-api.cc.
```
ehsan, line 950: FWIW in other tests that I have copied, I just included one line indicating the name of the original test.  I don't think this verbosity is needed!  I'll fix after merging.

mykmelez, line 950: Ah, I missed that the copyright header at the top of the file is "Copyright 2015 the V8 project authors." Ok, that makes sense, although I would then update the header to be "Copyright 2012 - 2015 the V8 project authors." (Or remove the year.)

ehsan, line 950: I don't think it matters, does it?

mykmelez, line 950: I'm unsure. I suggest it because test-api.cc is copyright 2012. Either way, the notice at the top of value.cc won't be an complete copyright header, just a pointer to other files. So maybe it doesn't matter.

Issue #51, opened 2016-05-03
Propagate the exception thrown from a Function call to the TryCatch on the caller's frame


ehsan:
We need to do something similar to what Chakrashim does for propagating exceptions to the caller's `TryCatch`, but `Function` is not implemented yet!

PR #55, opened 2016-05-04
Function

tbsaunde:

### DIFF
--- a/deps/spidershim/src/v8trycatch.cc
+++ b/deps/spidershim/src/v8trycatch.cc
@@ -41,7 +41,7 @@ struct TryCatch::Impl {
     // TODO: Propagate exceptions to the caller
     //       https://github.com/mozilla/spidernode/issues/51
   }
-  Isolate* Isolate() const { return isolate_; }
+  Isolate* Iso() const { return isolate_; }
   bool HasException() const {
     if (!hasExceptionSet_ &&
         !GetAndClearExceptionIfNeeded()) {
@@ -138,14 +138,14 @@ bool TryCatch::HasTerminated() const {
 
 Local<Value> TryCatch::ReThrow() {
   pimpl_->ReThrow();
-  return internal::Local<Value>::New(pimpl_->Isolate(), *pimpl_->Exception());
+  return internal::Local<Value>::New(pimpl_->Iso(), *pimpl_->Exception());
 }
 
 Local<Value> TryCatch::Exception() const {
   if (!pimpl_->HasException()) {
     return Local<Value>();
   }
-  return internal::Local<Value>::New(pimpl_->Isolate(), *pimpl_->Exception());
+  return internal::Local<Value>::New(pimpl_->Iso(), *pimpl_->Exception());
 }
 
 void TryCatch::Reset() {
--- a/deps/spidershim/test/value.cc
+++ b/deps/spidershim/test/value.cc
@@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
   Local<Float64Array> f64a =
       CreateAndCheck<Float64Array, 8>(buffer, 8, 127);
 }
+
+static inline v8::Local<v8::Script> v8_compile(v8::Local<v8::String> x) {
+  v8::Local<v8::Script> result;
+  if (v8::Script::Compile(v8::Isolate::GetCurrent()->GetCurrentContext(), x)
+          .ToLocal(&result)) {
+    return result;
+  }
+  return v8::Local<v8::Script>();
+}
+
+static inline v8::Local<v8::Value> CompileRun(v8::Local<v8::String> source) {
+  v8::Local<v8::Value> result;
+  if (v8_compile(source)
+          ->Run(v8::Isolate::GetCurrent()->GetCurrentContext())
+          .ToLocal(&result)) {
+    return result;
+  }
+  return v8::Local<v8::Value>();
+}
+
+
+// Helper functions that compile and run the source.
+static inline v8::Local<v8::Value> CompileRun(const char* source) {
+  return CompileRun(v8_str(source));
+}
+
+TEST(SpiderShim, FunctionCall) {
+  // This test is adopted from the V8 FunctionCall test.
+  V8Engine engine;
+
+  Isolate::Scope isolate_scope(engine.isolate());
+
+  HandleScope handle_scope(engine.isolate());
+  Local<Context> context = Context::New(engine.isolate());
+  Context::Scope context_scope(context);
+  Isolate* isolate = engine.isolate();
+
+  CompileRun(
+      "function Foo() {"
+      "  var result = [];"
+      "  for (var i = 0; i < arguments.length; i++) {"
+      "    result.push(arguments[i]);"
+      "  }"
+      "  return result;"
+      "}"
+      "function ReturnThisSloppy() {"
+      "  return this;"
+      "}"
+      "function ReturnThisStrict() {"
+      "  'use strict';"
+      "  return this;"
+      "}");
+  Local<Function> Foo = Local<Function>::Cast(
+      context->Global()->Get(context, v8_str("Foo")).ToLocalChecked());
+  Local<Function> ReturnThisSloppy = Local<Function>::Cast(
+      context->Global()
+          ->Get(context, v8_str("ReturnThisSloppy"))
+          .ToLocalChecked());
+  Local<Function> ReturnThisStrict = Local<Function>::Cast(
+      context->Global()
+          ->Get(context, v8_str("ReturnThisStrict"))
+          .ToLocalChecked());
+
+  v8::Local<Value>* args0 = NULL;
+  Local<v8::Array> a0 = Local<v8::Array>::Cast(
+      Foo->Call(context, Foo, 0, args0).ToLocalChecked());
+  EXPECT_EQ(0u, a0->Length());
+
+  v8::Local<Value> args1[] = {v8_num(1.1)};
+  Local<v8::Array> a1 = Local<v8::Array>::Cast(
+      Foo->Call(context, Foo, 1, args1).ToLocalChecked());
+  EXPECT_EQ(1u, a1->Length());
+  EXPECT_EQ(1.1, a1->Get(context, v8::Integer::New(isolate, 0))
+                    .ToLocalChecked()
+                    ->NumberValue(context)
+                    .FromJust());
+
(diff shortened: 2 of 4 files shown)


deps/spidershim/src/v8trycatch.cc
```diff
@@ -41,7 +41,7 @@ struct TryCatch::Impl {
     // TODO: Propagate exceptions to the caller
     //       https://github.com/mozilla/spidernode/issues/51
   }
-  Isolate* Isolate() const { return isolate_; }
+  Isolate* Iso() const { return isolate_; }
```
ehsan, line 44: Can you instead use `class Isolate* Isolate() const ...` instead?  That should help fix this error without needing to rename Isolate(), which is ugly.  We prefix names with `class` like this elsewhere in the code too.

deps/spidershim/test/value.cc
```diff
@@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
   Local<Float64Array> f64a =
       CreateAndCheck<Float64Array, 8>(buffer, 8, 127);
 }
+
+static inline v8::Local<v8::Script> v8_compile(v8::Local<v8::String> x) {
```
ehsan, line 1438: This is only used in `CompileRun` so it can be dropped.

deps/spidershim/test/value.cc
```diff
@@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
...
+  v8::Local<v8::Script> result;
+  if (v8::Script::Compile(v8::Isolate::GetCurrent()->GetCurrentContext(), x)
+          .ToLocal(&result)) {
+    return result;
+  }
+  return v8::Local<v8::Script>();
+}
+
+static inline v8::Local<v8::Value> CompileRun(v8::Local<v8::String> source) {
```
ehsan, line 1447: Can you instead please use `engine.CompileRun` like other tests do?

deps/spidershim/test/value.cc
```diff
@@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
...
+}
+
+
+// Helper functions that compile and run the source.
+static inline v8::Local<v8::Value> CompileRun(const char* source) {
+  return CompileRun(v8_str(source));
+}
+
+TEST(SpiderShim, FunctionCall) {
+  // This test is adopted from the V8 FunctionCall test.
```
ehsan, line 1464: Awesome!!!

deps/spidershim/test/value.cc
```diff
@@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
...
+          .ToLocalChecked();
+  EXPECT_TRUE(r8->StrictEquals(v8_num(42)));
+  Local<v8::Value> r9 =
+      ReturnThisStrict->Call(context, v8_str("hello"), 0, NULL)
+          .ToLocalChecked();
+  EXPECT_TRUE(r9->StrictEquals(v8_str("hello")));
+  Local<v8::Value> r10 =
+      ReturnThisStrict->Call(context, v8::True(isolate), 0, NULL)
+          .ToLocalChecked();
+  EXPECT_TRUE(r10->StrictEquals(v8::True(isolate)));
```
ehsan, line 1609: FWIW, all of these `v8::` prefixes are unnecessary.  :)

(I can take them out after merging too.)

deps/spidershim/src/v8trycatch.cc
tbsaunde, line 44: On Wed, May 04, 2016 at 10:39:55PM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8trycatch.cc:
> 
> > @@ -41,7 +41,7 @@ struct TryCatch::Impl {
> >      // TODO: Propagate exceptions to the caller
> >      //       https://github.com/mozilla/spidernode/issues/51
> >    }
> > -  Isolate\* Isolate() const { return isolate_; }
> > -  Isolate\* Iso() const { return isolate_; }
> 
> Can you instead use class Isolate\* Isolate() const ... instead? That should
> help fix this error without needing to rename Isolate(), which is ugly. We
> prefix names with class like this elsewhere in the code too.

No, that doesn't work because the method is still changing the meaning
of Isolate in the class scope from the type to the method, which aiui is
the problem.  Its also what Botond suggested, and I'm not going to try
and question him on C++ name lookup rules ;)

deps/spidershim/test/value.cc
tbsaunde, line 1447: On Wed, May 04, 2016 at 10:40:48PM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/test/value.cc:
> 
> > @@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
> >    Local<Float64Array> f64a =
> >        CreateAndCheck<Float64Array, 8>(buffer, 8, 127);
> >  }
> > +
> > +static inline v8::Localv8::Script v8_compile(v8::Localv8::String x) {
> > -  v8::Localv8::Script result;
> > -  if (v8::Script::Compile(v8::Isolate::GetCurrent()->GetCurrentContext(), x)
> > -          .ToLocal(&result)) {
> > -    return result;
> > -  }
> > -  return v8::Localv8::Script();
> >   +}
> >   +
> >   +static inline v8::Localv8::Value CompileRun(v8::Localv8::String source) {
> 
> Can you instead please use engine.CompileRun like other tests do?

I could try I guess, but is there really value in diverging from the v8
tests?  I think there may well be more v8 tests using these helpers, and
I don't really see a reason to make stealing them harder.

deps/spidershim/src/v8trycatch.cc
ehsan, line 44: OK, I'll fix this later myself.  What version of gcc did you get this error with?

deps/spidershim/test/value.cc
ehsan, line 1447: I don't think it will make anything harder, it will make your test consistent with the existing ones.  I'll do this myself, no point debating this.  :-)

ehsan, line 1447: But I'd appreciate if you try to make the tests more consistent in future PRs.  Thanks!

deps/spidershim/src/v8trycatch.cc
tbsaunde, line 44: On Thu, May 05, 2016 at 11:42:54AM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8trycatch.cc:
> 
> > @@ -41,7 +41,7 @@ struct TryCatch::Impl {
> >      // TODO: Propagate exceptions to the caller
> >      //       https://github.com/mozilla/spidernode/issues/51
> >    }
> > -  Isolate\* Isolate() const { return isolate_; }
> > -  Isolate\* Iso() const { return isolate_; }
> 
> OK, I'll fix this later myself. What version of gcc did you get this error
> with?

I think everyone on my machine so 4.9 / 5 / 6.  At least with a reduced
test case I used to figure out what was going on.

deps/spidershim/test/value.cc
tbsaunde, line 1447: On Thu, May 05, 2016 at 11:44:44AM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/test/value.cc:
> 
> > @@ -1434,3 +1434,177 @@ TEST(SpiderShim, ArrayBuffer_NeuteringApi) {
> >    Local<Float64Array> f64a =
> >        CreateAndCheck<Float64Array, 8>(buffer, 8, 127);
> >  }
> > +
> > +static inline v8::Localv8::Script v8_compile(v8::Localv8::String x) {
> > -  v8::Localv8::Script result;
> > -  if (v8::Script::Compile(v8::Isolate::GetCurrent()->GetCurrentContext(), x)
> > -          .ToLocal(&result)) {
> > -    return result;
> > -  }
> > -  return v8::Localv8::Script();
> >   +}
> >   +
> >   +static inline v8::Localv8::Value CompileRun(v8::Localv8::String source) {
> 
> I don't think it will make anything harder, it will make your test consistent
> with the existing ones. I'll do this myself, no point debating this. :-)

The more different our tests are than the v8 ones the more work you need
to do to import them, but if you insist I'll try and remember to do all
the adjustments to use the local style.

deps/spidershim/src/v8trycatch.cc
ehsan, line 44: Do you mind pasting the test case here?  This program builds fine with gcc5 for me:

``` cpp
class Foo;
class Bar {
  class Foo* Foo();
};
```

deps/spidershim/test/value.cc
ehsan, line 1447: Well, for now the tests cannot be imported verbatim anyways, and given that our `V8Engine` setup is different than `cctest`'s, I'd rather keep the tests calling the `V8Engine` `CompileRun` method.  FWIW this is fixed on master.

deps/spidershim/src/v8trycatch.cc
tbsaunde, line 44: On Thu, May 05, 2016 at 12:21:27PM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8trycatch.cc:
> 
> > @@ -41,7 +41,7 @@ struct TryCatch::Impl {
> >      // TODO: Propagate exceptions to the caller
> >      //       https://github.com/mozilla/spidernode/issues/51
> >    }
> > -  Isolate\* Isolate() const { return isolate_; }
> > -  Isolate\* Iso() const { return isolate_; }
> 
> Do you mind pasting the test case here? This program builds fine with gcc5 for
> me:
> 
> class Foo;
> class Bar {
>   class Foo\* Foo();
> };

I was really suprised that works for you, so I poked at it quickly.
Indeed that is fine (I'm not sure what the spec says but it compiles).
However this does not

class Isolate {};
class Foo {
    Isolate\* isolate_;
    public:
    class Isolate\* Isolate() const;
};

Which makes sense mod my not understanding name lookup, the type of
Isolate_ could be unclear.  However this is fine:

class Isolate {};
class Foo {
    class Isolate\* isolate_;
    public:
    class Isolate\* Isolate() const;
};

So I guess if you make it clear all over you mean the class it works?
I'm not sure where all you'd need to add class.

ehsan, line 44: In this case, the `isolate_` member is already declared as `class Isolate*`, so I think everything would be fine here with my original suggestion.  I did that in 9176b11.  If that breaks the build for you, please file an issue.  Thanks!

PR #56, opened 2016-05-05
implement String interning

mykmelez:
This branch implements string interning by using JS_AtomizeAndPinUCString to intern the JSString objects created by the various String::New\* APIs that take a NewStringType.

Note that V8 conserves the String objects for interned strings, whereas in this implementation SpiderShim creates new String objects that conserve the JSString objects they wrap. I don't think this implementation detail affects API compatibility, however.

### DIFF
--- a/deps/spidershim/src/v8string.cc
+++ b/deps/spidershim/src/v8string.cc
@@ -77,7 +77,6 @@ Local<String> String::NewFromUtf8(Isolate* isolate, const char* data,
 
 MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
                                        v8::NewStringType type, int length) {
-  assert(type == v8::NewStringType::kNormal); // TODO: Add support for interned strings
   JSContext* cx = JSContextFromIsolate(isolate);
 
   // In V8's api.cc, this conditional block is annotated with the comment:
@@ -96,27 +95,40 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
   if (!twoByteChars) {
     return MaybeLocal<String>();
   }
-  JS::RootedString str(cx, JS_NewUCString(cx, twoByteChars.release(), twoByteLen));
+
+  JSString* str =
+    type == v8::NewStringType::kInternalized ?
+      JS_AtomizeAndPinUCString(cx, reinterpret_cast<const char16_t*>(twoByteChars.release())) :
+      JS_NewUCString(cx, twoByteChars.release(), twoByteLen);
+
   if (!str) {
     return MaybeLocal<String>();
   }
+
+  JS::RootedString rootedStr(cx, str);
   JS::Value strVal;
-  strVal.setString(str);
+  strVal.setString(rootedStr);
   return internal::Local<String>::New(isolate, strVal);
 }
 
 MaybeLocal<String> String::NewFromOneByte(Isolate* isolate, const uint8_t* data,
                                           v8::NewStringType type, int length) {
-  assert(type == v8::NewStringType::kNormal); // TODO: Add support for interned strings
   JSContext* cx = JSContextFromIsolate(isolate);
-  JS::RootedString str(cx, length >= 0 ?
-    JS_NewStringCopyN(cx, reinterpret_cast<const char*>(data), length) :
-    JS_NewStringCopyZ(cx, reinterpret_cast<const char*>(data)));
+
+  JSString* str =
+    type == v8::NewStringType::kInternalized ?
+      JS_AtomizeAndPinString(cx, reinterpret_cast<const char*>(data)) :
+      length >= 0 ?
+        JS_NewStringCopyN(cx, reinterpret_cast<const char*>(data), length) :
+        JS_NewStringCopyZ(cx, reinterpret_cast<const char*>(data));
+
   if (!str) {
     return MaybeLocal<String>();
   }
+
+  JS::RootedString rootedStr(cx, str);
   JS::Value strVal;
-  strVal.setString(str);
+  strVal.setString(rootedStr);
   return internal::Local<String>::New(isolate, strVal);
 }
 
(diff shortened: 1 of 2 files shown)


deps/spidershim/src/v8string.cc
```diff
@@ -96,27 +95,40 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
   if (!twoByteChars) {
     return MaybeLocal<String>();
   }
-  JS::RootedString str(cx, JS_NewUCString(cx, twoByteChars.release(), twoByteLen));
+
+  JSString* str =
+    type == v8::NewStringType::kInternalized ?
+      JS_AtomizeAndPinUCString(cx, reinterpret_cast<const char16_t*>(twoByteChars.release())) :
```
ehsan, line 101: You're losing the length here, right?  I think you should use `JS_AtomizeAndPinUCStringN` instead.  Can you also add a test case that would have caught this please?

deps/spidershim/src/v8string.cc
```diff
@@ -96,27 +95,40 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
...
+      JS_NewUCString(cx, twoByteChars.release(), twoByteLen);
+
   if (!str) {
     return MaybeLocal<String>();
   }
+
+  JS::RootedString rootedStr(cx, str);
```
ehsan: Why do you need to root the string here?  The `internal::Local<String>::New()` call below does that for you, and nothing can GC here between these two points.

deps/spidershim/src/v8string.cc
```diff
@@ -96,27 +95,40 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
...
                                           v8::NewStringType type, int length) {
-  assert(type == v8::NewStringType::kNormal); // TODO: Add support for interned strings
   JSContext* cx = JSContextFromIsolate(isolate);
-  JS::RootedString str(cx, length >= 0 ?
-    JS_NewStringCopyN(cx, reinterpret_cast<const char*>(data), length) :
-    JS_NewStringCopyZ(cx, reinterpret_cast<const char*>(data)));
+
+  JSString* str =
+    type == v8::NewStringType::kInternalized ?
+      JS_AtomizeAndPinString(cx, reinterpret_cast<const char*>(data)) :
```
ehsan: Same comment about the length here.

deps/spidershim/src/v8string.cc
```diff
@@ -96,27 +95,40 @@ MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
...
+      length >= 0 ?
+        JS_NewStringCopyN(cx, reinterpret_cast<const char*>(data), length) :
+        JS_NewStringCopyZ(cx, reinterpret_cast<const char*>(data));
+
   if (!str) {
     return MaybeLocal<String>();
   }
+
+  JS::RootedString rootedStr(cx, str);
```
ehsan: Same here too.

deps/spidershim/src/v8string.cc
```diff
@@ -128,16 +140,22 @@ Local<String> String::NewFromOneByte(Isolate* isolate, const uint8_t* data,
...
                                           v8::NewStringType type, int length) {
-  assert(type == v8::NewStringType::kNormal); // TODO: Add support for interned strings
   JSContext* cx = JSContextFromIsolate(isolate);
-  JS::RootedString str(cx, length >= 0 ?
-    JS_NewUCStringCopyN(cx, reinterpret_cast<const char16_t*>(data), length) :
-    JS_NewUCStringCopyZ(cx, reinterpret_cast<const char16_t*>(data)));
+
+  JSString* str =
+    type == v8::NewStringType::kInternalized ?
+      JS_AtomizeAndPinUCString(cx, reinterpret_cast<const char16_t*>(data)) :
```
ehsan: And here.

deps/spidershim/src/v8string.cc
```diff
@@ -128,16 +140,22 @@ Local<String> String::NewFromOneByte(Isolate* isolate, const uint8_t* data,
...
+      length >= 0 ?
+        JS_NewUCStringCopyN(cx, reinterpret_cast<const char16_t*>(data), length) :
+        JS_NewUCStringCopyZ(cx, reinterpret_cast<const char16_t*>(data));
+
   if (!str) {
     return MaybeLocal<String>();
   }
+
+  JS::RootedString rootedStr(cx, str);
```
ehsan: And here too!

deps/spidershim/test/value.cc
```diff
@@ -32,8 +32,22 @@ static int StrNCmp16(uint16_t* a, uint16_t* b, int n) {
   }
 }
 
+// Based on the function of the same name in test-api.cc, but modified
+// to compare strings in a SpiderShim-compatible way.
+static bool SameSymbol(Local<String> s1, Local<String> s2) {
+  Isolate* isolate = Isolate::GetCurrent();
+  // Use Persistent<T> to get access to the protected Value "equal to" operator.
```
ehsan, line 53: Hmm, I'm not sure if I understand the assumption behind this code.  Can you please explain what this is trying to ensure?  It's probably worth clarifying the comment here...

deps/spidershim/src/v8string.cc
mykmelez: > Why do you need to root the string here? The internal::Local<String>::New() call below does that for you, and nothing can GC here between these two points.

This is the existing behavior, as you originally implemented it, and I'm not changing it in this PR.

mykmelez: By "I'm not changing it," I mean "I haven't changed it," of course. I'm happy to make the change if the existing behavior is no longer (or never was) the correct behavior.

ehsan: Oh hah right!  I missed that you're just moving the code.  :-)

I'm not sure why I've done this, looks like I was over-zealous.  I'm happy to fix this myself after merging!

deps/spidershim/src/v8string.cc
mykmelez, line 101: > You're losing the length here, right? I think you should use JS_AtomizeAndPinUCStringN instead. Can you also add a test case that would have caught this please?

Indeed, I completely forgot about that. Fixed in 76e04ac, and I added tests in fcd2689.

mykmelez, line 101: Erm, except that the fix is just for String::NewFromOneByte and String::NewFromTwoByte. String::NewFromUtf8 isn't susceptible to the bug, since it calls JS::UTF8CharsToNewTwoByteCharsZ, passing it the length, to get an already-truncated JS::UniqueTwoByteChars that it can pass to JS_AtomizeAndPinUCString without having to specify a length.

deps/spidershim/src/v8string.cc
mykmelez: > I'm not sure why I've done this, looks like I was over-zealous. I'm happy to fix this myself after merging!

No worries, I've done so in 1aa10f6.

deps/spidershim/test/value.cc
mykmelez, line 53: > Hmm, I'm not sure if I understand the assumption behind this code. Can you please explain what this is trying to ensure? It's probably worth clarifying the comment here...

It ensures that the JSString object referenced by the first Local<String> is the same JSString object referenced by the second Local<String>, i.e. that both Local<String>s reference the same JSString (as they should do when the strings were interned) rather then each referencing a different JSString (as they should do when the strings weren't interned).

I've expanded the commentary (and added a test to confirm that SameSymbol returns false for non-interned strings) in 99faddb.

mykmelez, line 53: (Of course it's also possible that my assumptions about Value and equivalence are wrong, even though the tests pass. If so, please enlighten me!)

PR #57, opened 2016-05-05
spidershim: Implement and stub out some of the v8::V8 methods

### LINKED ISSUE #31: Finish implementing v8::V8
The following methods are used by Node and yet unimplemented:

```
  v8::V8::GetVersion()
  v8::V8::SetEntropySource(bool (*)(unsigned char*, unsigned long))
  v8::V8::SetFlagsFromCommandLine(int*, char**, bool)
  v8::V8::SetFlagsFromString(char const*, int)
```

brendandahl:
Fixes #31

I implemented the trivial version one, the rest of these seem low priority for getting up and running.
- GetVersion - may need some work to make the version number similar to a node version number
- SetFlagsFrom\* - I looked into fully implementing this by using jsshell's argument parser and option setter, but it's built into the main, so if we want to use that we'll need to do some work to extract it
- SetEntropySource - also stubbed out in chakra

### DIFF
--- a/deps/spidershim/src/v8v8.cc
+++ b/deps/spidershim/src/v8v8.cc
@@ -55,21 +55,19 @@ bool V8::IsDead() {
 }
 
 const char *V8::GetVersion() {
-  // Spidermonkey's version number does not match the format of v8's (X.X.X.X),
-  // unsure if this will be an issue yet.
   return JS_GetImplementationVersion();
 }
 
 void V8::SetFlagsFromString(const char* str, int length) {
-  // TODO
+  // TODO: see SetFlagsFromCommandLine
 }
 
 void V8::SetFlagsFromCommandLine(int *argc, char **argv, bool remove_flags) {
-  // TODO
+  // TODO: command line arguments should be added on an as-needed basis
 }
 
 void V8::SetEntropySource(EntropySource entropy_source) {
-  // TODO
+  // TODO: https://github.com/mozilla/spidernode/issues/58
 }
 
 }
--- a/deps/spidershim/src/v8v8.cc
+++ b/deps/spidershim/src/v8v8.cc
@@ -54,4 +54,22 @@ bool V8::IsDead() {
   return internal::gDisposed;
 }
 
+const char *V8::GetVersion() {
+  // Spidermonkey's version number does not match the format of v8's (X.X.X.X),
+  // unsure if this will be an issue yet.
+  return JS_GetImplementationVersion();
+}
+
+void V8::SetFlagsFromString(const char* str, int length) {
+  // TODO
+}
+
+void V8::SetFlagsFromCommandLine(int *argc, char **argv, bool remove_flags) {
+  // TODO
+}
+
+void V8::SetEntropySource(EntropySource entropy_source) {
+  // TODO
+}
+
 }
--- a/deps/spidershim/test/v8.cc
+++ b/deps/spidershim/test/v8.cc
@@ -11,5 +11,5 @@
 #include "gtest/gtest.h"
 
 TEST(SpiderShim, Version) {
-  v8::V8::GetVersion();
+  EXPECT_EQ(strncmp("JavaScript-C", v8::V8::GetVersion(), 12), 0);
 }
--- a/deps/spidershim/test/v8v8.cc
+++ b/deps/spidershim/test/v8v8.cc
@@ -1,0 +1,15 @@
+// Copyright 2015 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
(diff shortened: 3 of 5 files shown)


deps/spidershim/src/v8v8.cc
```diff
@@ -54,4 +54,22 @@ bool V8::IsDead() {
   return internal::gDisposed;
 }
 
+const char *V8::GetVersion() {
+  // Spidermonkey's version number does not match the format of v8's (X.X.X.X),
+  // unsure if this will be an issue yet.
+  return JS_GetImplementationVersion();
```
ehsan, line 58: This is fine!  Node [exposes](https://github.com/mozilla/spidernode/blob/312cf60c5229b50c386c56f5f993a3c6717ee18e/src/node.cc#L2934) this information as the version of the backend JS engine.

deps/spidershim/src/v8v8.cc
```diff
@@ -54,4 +54,22 @@ bool V8::IsDead() {
...
+}
+
+void V8::SetFlagsFromString(const char* str, int length) {
+  // TODO
+}
+
+void V8::SetFlagsFromCommandLine(int *argc, char **argv, bool remove_flags) {
+  // TODO
```
ehsan: These two will probably be difficult to implement, and we should accept command line arguments on an as-needed basis.  Do you mind adding a comment to that effect please?

deps/spidershim/src/v8v8.cc
```diff
@@ -54,4 +54,22 @@ bool V8::IsDead() {
...
+}
+
+void V8::SetEntropySource(EntropySource entropy_source) {
+  // TODO
```
ehsan: SpiderMonkey doesn't support setting entropy sources, but we should test the original issue that caused this call to be added to Node to ensure that we won't suffer from a similar issue.  Filed #58 to track this.  Can you please add a comment pointing to that issue?

deps/spidershim/test/v8v8.cc
```diff
@@ -0,0 +1,15 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "v8engine.h"
+
+#include "gtest/gtest.h"
+
+TEST(SpiderShim, Version) {
+  v8::V8::GetVersion();
```
ehsan: Why not make this actually test something?  Like compare it against `"JavaScript-C<version>" or something like that using`strncmp()`?

deps/spidershim/tests.gyp
```diff
@@ -28,5 +28,9 @@
       'target_name': 'value',
       'sources': [ 'test/value.cc' ],
     },
+    {
+      'target_name': 'v8v8',
+      'sources': [ 'test/v8v8.cc' ],
```
ehsan: Please add this to [run-tests.sh](https://github.com/mozilla/spidernode/blob/master/deps/spidershim/scripts/run-tests.sh#L12) too.  (We don't have a fully functioning build system yet, so we need to manually get the tests to run on Travis-CI!)

deps/spidershim/tests.gyp
```diff
@@ -28,5 +28,9 @@
       'target_name': 'value',
       'sources': [ 'test/value.cc' ],
     },
+    {
+      'target_name': 'v8v8',
```
ehsan: Nit: For consistency, please rename this to `v8` and `v8.cc`.  We don't use the "v8" prefix for test file names.

ehsan:
Thanks for the PR!  It mostly looks good, but I'd appreciate if you can please address my comments above too before I merge.

brendandahl:
I'll update. Not sure if you saw, but some these things I talk about in the PR description.

deps/spidershim/test/v8v8.cc
brendandahl: I just used the test from test-api.cc and wasn't sure if this string is stable. I'll add a further test.

ehsan: Seems to be [stable](https://github.com/mozilla/spidernode/blob/a01a274ef82b589fc846b686e1f243fc47b92df3/deps/spidershim/spidermonkey/js/src/jsapi.cpp#L712) now.  We can change the test later if this changes in the future...

ehsan:
> I'll update. Not sure if you saw, but some these things I talk about in the PR description.

Oh right yeah I think I missed that! :)

deps/spidershim/test/v8.cc
```diff
@@ -11,5 +11,5 @@
 #include "gtest/gtest.h"
 
 TEST(SpiderShim, Version) {
-  v8::V8::GetVersion();
+  EXPECT_EQ(strncmp("JavaScript-C", v8::V8::GetVersion(), 12), 0);
```
ehsan, line 14: The order of arguments is the reverse (expected first).  I'll fix after merging.

ehsan:
Thanks a lot!

PR #59, opened 2016-05-06
add ScriptUsing*StringResource tests

mykmelez:
This branch adapts two of the ExternalStringResource tests in test-api.cc to tests in value.cc. Most of the test assertions are commented out, because SpiderShim doesn't yet support the features being tested (or the test harness doesn't support them), like String::IsExternal and String::GetExternalStringResourceBase.

So I'm not sure how useful these tests are, nor how useful it would be to adapt more of the ExternalStringResource tests in test-api.cc. Of course we could always implement those features, but since Node doesn't depend on them, they don't seem like a high priority.

### DIFF
--- a/deps/spidershim/test/v8engine.h
+++ b/deps/spidershim/test/v8engine.h
@@ -95,6 +95,14 @@ public:
     return Local<Value>();
   }
 
+  Local<Value> CompileRun(Local<Context> context, Local<String>* script) {
+    auto scr = Script::Compile(*script);
+    if (*scr) {
+      return scr->Run();
+    }
+    return Local<Value>();
+  }
+
 private:
   Isolate* isolate_;
 };
--- a/deps/spidershim/test/v8engine.h
+++ b/deps/spidershim/test/v8engine.h
@@ -95,7 +95,7 @@ public:
     return Local<Value>();
   }
 
-  Local<Value> CompileRun(Local<Context> context, Local<String> script) {
+  Local<Value> CompileRun(Local<String> script) {
     auto scr = Script::Compile(script);
     if (*scr) {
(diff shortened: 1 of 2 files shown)


ehsan:
> So I'm not sure how useful these tests are, nor how useful it would be to adapt more of the ExternalStringResource tests in test-api.cc. Of course we could always implement those features, but since Node doesn't depend on them, they don't seem like a high priority.

Agreed!

deps/spidershim/test/v8engine.h
```diff
@@ -95,6 +95,14 @@ class V8Engine {
     return Local<Value>();
   }
 
+  Local<Value> CompileRun(Local<Context> context, Local<String>* script) {
```
ehsan: Passing `Local` by pointer doesn't make sense.  Did you mean to pass it by value?

Also it would be nice if you refactored the other `CompileRun` overload to be implemented in terms of this using `v8_str` or something like that...

ehsan:
Looks good except for the one comment I posted on `CompileRun`...

mykmelez:
> Looks good except for the one comment I posted on CompileRun...

Fixed in 5209375. In the process, I realized that the _context_ parameter is unused, so I removed it in d37915d. (That's also true for the CompileRun variant that takes a `const char* script`, which originally used that parameter but doesn't anymore.)

deps/spidershim/test/v8engine.h
mykmelez: > Also it would be nice if you refactored the other CompileRun overload to be implemented in terms of this using v8_str or something like that...

Yup, makes sense, done in eaa88ed.

PR #62, opened 2016-05-06
spidershim: Implement Function::NewInstance

brendandahl:


deps/spidershim/src/v8function.cc
```diff
@@ -27,6 +27,32 @@
...
+    return Local<Object>();
+  }
+
+  for (int i = 0; i < argc; i++) {
+    args.infallibleAppend(*reinterpret_cast<JS::Value*>(*argv[i]));
+  }
+
+  auto obj = ::JS_New(cx, thisObj, args);
+  if (!obj) {
+    return Local<Object>();
```
ehsan: Nit: `MaybeLocal<Object>()`

deps/spidershim/test/value.cc
```diff
@@ -1870,6 +1870,94 @@ TEST(SpiderShim, FunctionCall) {
...
+      "  var result = [];"
+      "  for (var i = 0; i < arguments.length; i++) {"
+      "    result.push(arguments[i]);"
+      "  }"
+      "  return result;"
+      "}");
+  Local<Function> Foo = Local<Function>::Cast(
+      context->Global()->Get(context, v8_str("Foo")).ToLocalChecked());
+
+  v8::Local<Value>* args0 = NULL;
```
ehsan: Can you please remove these needless `v8::` prefixes?  I find the tests much easier to read without this spread all over them.  :-)

ehsan:
Thanks, looks great!  I have a couple of nits, do you mind addressing them please before I merge?

deps/spidershim/src/v8function.cc
```diff
@@ -27,6 +27,32 @@
 
 namespace v8 {
 
+MaybeLocal<Object>
+Function::NewInstance(Local<Context> context,
+                      int argc, Handle<Value> argv[]) const
+{
+  Isolate* isolate = Isolate::GetCurrent();
+  JSContext* cx = JSContextFromIsolate(isolate);
+  JS::RootedObject thisObj(cx, &reinterpret_cast<const JS::Value*>(this)->toObject());
```
mykmelez, line 36: @ehsan Is it also unnecessary to root this here, because the internal::Local<Object>::New() constructor will do that?

brendandahl, line 36: JS_New requires a handle, is there some other way to get a handle here?

mykmelez, line 36: > JS_New requires a handle, is there some other way to get a handle here?

Ah, I missed the JS_New call. I don't know of another way to get a handle here, so never mind. ;-)

ehsan, line 36: > JS_New requires a handle, is there some other way to get a handle here?
> 
> Ah, I missed the JS_New call. I don't know of another way to get a handle here, so never mind. ;-)

Yeah, this is the reason for a lot of things in spidershim to be Rooted where they didn't need to except because of this...

PR #63, opened 2016-05-06
ensure string length <= max length

mykmelez:
Only String::NewFromUtf8 is currently checking that the new string data's length is <= String::kMaxLength. And String::Concat doesn't check that the concatenation is within the upper bound either. All the String::New\* methods and String::Concat should have this check. This branch adds it, plus tests.

### DIFF
--- a/deps/spidershim/src/v8string.cc
+++ b/deps/spidershim/src/v8string.cc
@@ -77,17 +77,17 @@ Local<String> String::NewFromUtf8(Isolate* isolate, const char* data,
 
 MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, const char* data,
                                        v8::NewStringType type, int length) {
-  JSContext* cx = JSContextFromIsolate(isolate);
-
   // In V8's api.cc, this conditional block is annotated with the comment:
   //    TODO(dcarney): throw a context free exception.
-  if (length > v8::String::kMaxLength) {
+  if (length > String::kMaxLength) {
     return MaybeLocal<String>();
   }
 
   if (length < 0) {
     length = strlen(data);
   }
+
+  JSContext* cx = JSContextFromIsolate(isolate);
 
   size_t twoByteLen;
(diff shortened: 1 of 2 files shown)


mykmelez:
Hmm, StringConcatOverflow segfaults on Travis (although not locally). I'm not sure what to do about that.

ehsan:
> Hmm, StringConcatOverflow segfaults on Travis (although not locally). I'm not sure what to do about that.

I'm not sure if I understand the test yet.  What is supposed to initialize `RandomLengthOneByteResource::string_`?  As far as I can tell this test may end up using some random memory...

mykmelez:
> I'm not sure if I understand the test yet. What is supposed to initialize RandomLengthOneByteResource::string_? As far as I can tell this test may end up using some random memory...

I was puzzling over that too. In the original test, the memory is never initialized:

https://github.com/mozilla/spidernode/blob/master/deps/v8/test/cctest/test-api.cc#L23979-L23991

Nor is it initialized in the only other test that uses RandomLengthOneByteResource:

https://github.com/mozilla/spidernode/blob/master/deps/v8/test/cctest/test-api.cc#L727-L748

I suspect the idea is that it doesn't need to be initialized, because neither String::NewExternalOneByte nor String::Concat actually access the buffer (since NewExternalOneByte doesn't need to access it, and Concat exits early after determining that the concatenated string would be too long).

Ah, but our implementation of String::NewExternalOneByte actually copies the data:

https://github.com/mozilla/spidernode/blob/master/deps/spidershim/src/v8string.cc#L198-L212

ehsan:
Yeah, that should be the issue.  I expect the crash to go away once you initialize the buffer.

mykmelez:
> Yeah, that should be the issue. I expect the crash to go away once you initialize the buffer.

Indeed, I made that change in 5b75fea, then tested it on Travis in my fork, and the crash went away there.

PR #67, opened 2016-05-07
enable Travis for all branches

mykmelez:
@ehsan Is it important to disable Travis on branches? There's only one other branch in the repository, container-based, and it appears to be defunct. Enabling Travis for branches makes it easier to test changes on Travis in my fork, since otherwise I have to enable it on each branch manually by modifying .travis.yml (and then undo those changes before requesting a pull).


ehsan:
I don't know why I had this set, I blame cargo-culting.  Thanks for the fix!  :-)

PR #68, opened 2016-05-10
Array

tbsaunde:
This is mostly a place holder that I got this done, commits need some clean up, and I need to figure out what tests can be ported.  I may also try and get the spidermonkey patch in we'll see how that goes.

### DIFF
--- a/deps/spidershim/spidermonkey-patches/02-NewArrayBufferWithExternalContents.patch
+++ b/deps/spidershim/spidermonkey-patches/02-NewArrayBufferWithExternalContents.patch
@@ -1,0 +1,39 @@
+diff --git a/deps/spidershim/spidermonkey/js/src/jsapi.h b/deps/spidershim/spidermonkey/js/src/jsapi.h
+index 5ae71a7..91ad900 100644
+--- a/deps/spidershim/spidermonkey/js/src/jsapi.h
++++ b/deps/spidershim/spidermonkey/js/src/jsapi.h
+@@ -3500,6 +3500,13 @@ extern JS_PUBLIC_API(JSObject*)
+ JS_NewArrayBufferWithContents(JSContext* cx, size_t nbytes, void* contents);
+ 
+ /**
++ * Create a new array buffer with the given contents.  The array buffer does not take ownership of
++ * contents, and JS_DetachArrayBuffer must be called befor the contents are disposed of.
++ */
++extern JS_PUBLIC_API(JSObject*)
++JS_NewArrayBufferWithExternalContents(JSContext* cx, size_t nbytes, void* contents);
++
++/**
+  * Steal the contents of the given array buffer. The array buffer has its
+  * length set to 0 and its contents array cleared. The caller takes ownership
+  * of the return value and must free it or transfer ownership via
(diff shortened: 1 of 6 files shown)


deps/spidershim/spidermonkey-patches/02-NewArrayBufferWithExternalContents.patch
```diff
@@ -0,0 +1,39 @@
+diff --git a/deps/spidershim/spidermonkey/js/src/jsapi.h b/deps/spidershim/spidermonkey/js/src/jsapi.h
```
ehsan, line 1: Can you please upstream this patch so that we don't have to maintain it in spidershim?  (Obviously landing it downstream in the mean time is fine!)

deps/spidershim/src/v8arraybuffer.cc
```diff
@@ -122,6 +122,14 @@ ArrayBufferView::ByteLength()
   return JS_GetDataViewByteLength(view);
 }
 
+void
+ArrayBuffer::Neuter()
+{
+  JSContext* cx = JSContextFromIsolate(Isolate::GetCurrent());
+  JS::RootedObject obj(cx, &reinterpret_cast<JS::Value*>(this)->toObject());
+  JS_DetachArrayBuffer(cx, obj, KeepData);
```
ehsan, line 130: Hmm, what happens to the underlying buffer at this stage?  As far as I can tell, we'd be leaking it here...

deps/spidershim/spidermonkey-patches/02-NewArrayBufferWithExternalContents.patch
tbsaunde, line 1: On Tue, May 10, 2016 at 09:57:11AM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/spidermonkey-patches/
> 02-NewArrayBufferWithExternalContents.patch:
> 
> > @@ -0,0 +1,39 @@
> > +diff --git a/deps/spidershim/spidermonkey/js/src/jsapi.h b/deps/spidershim/spidermonkey/js/src/jsapi.h
> 
> Can you please upstream this patch so that we don't have to maintain it in
> spidershim? (Obviously landing it downstream in the mean time is fine!)

I was already doing that

deps/spidershim/src/v8arraybuffer.cc
tbsaunde, line 130: On Tue, May 10, 2016 at 10:08:32AM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8arraybuffer.cc:
> 
> > @@ -122,6 +122,14 @@ ArrayBufferView::ByteLength()
> >    return JS_GetDataViewByteLength(view);
> >  }
> > 
> > +void
> > +ArrayBuffer::Neuter()
> > +{
> > -  JSContext\* cx = JSContextFromIsolate(Isolate::GetCurrent());
> > -  JS::RootedObject obj(cx, &reinterpret_castJS::Value*(this)->toObject());
> > -  JS_DetachArrayBuffer(cx, obj, KeepData);
> 
> Hmm, what happens to the underlying buffer at this stage? As far as I can tell,
> we'd be leaking it here...

If js owns the data then JS_DetachArrayBuffer() will free it.  However
in v8 it seems to be valid  only to call this on ArrayBuffer's where the
embedder owns the data.  So in the places this should be called it is
intended that the buffer doesn't own the data and so not freeing it is
correct.

PR #82, opened 2016-05-11
spidershim: implement out v8::Isolate::<Add,Remove>GC<Epilogue,Prologue>Ca…

brendandahl:
follow up in #107

### DIFF
--- a/deps/spidershim/src/v8isolate.cc
+++ b/deps/spidershim/src/v8isolate.cc
@@ -28,9 +28,30 @@
 #include "mozilla/TimeStamp.h"
 #include "mozilla/ThreadLocal.h"
 
-namespace {
+namespace v8 {
 
-void OnGC(JSRuntime* rt, JSGCStatus status, void* data) {
+static MOZ_THREAD_LOCAL(Isolate*) sCurrentIsolate;
+
+namespace internal {
+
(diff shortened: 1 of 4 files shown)


deps/spidershim/src/v8isolate.cc
```diff
@@ -193,6 +193,22 @@ void Isolate::AddStackTrace(StackTrace* trace) {
   pimpl_->stackTraces.push_back(trace);
 }
 
+void Isolate::AddGCPrologueCallback(
+  GCCallback callback, GCType gc_type_filter) {
+  // TODO: used only with dtrace/etw, could be implemented with JS_SetGCCallback
```
ehsan: @brendandahl given that we already have `OnGC()` in `v8isolate.cc`, any reason why we wouldn't just hook these up for reals? :)

ehsan:
Ping?

deps/spidershim/src/v8isolate.cc
brendandahl: > @brendandahl given that we already have OnGC() in v8isolate.cc, any reason why we wouldn't just hook these up for reals? :)

I'll look into hooking them up now. I've been mainly stubbing out what chakra got away with.

brendandahl:
@ehsan ready for review

ehsan:
Looks great, thanks!

PR #86, opened 2016-05-12
Embed

tbsaunde:

### DIFF
--- a/deps/spidershim/src/conversions.h
+++ b/deps/spidershim/src/conversions.h
@@ -40,6 +40,10 @@ static inline Value* GetV8Value(JS::Value* val) {
   return reinterpret_cast<Value*>(val);
 }
 
+static inline Value* GetV8Value(JS::MutableHandleValue val) {
+  return GetV8Value(val.address());
+}
+
 static inline JSObject* GetObject(Value* val) {
   JS::Value* v = GetValue(val);
   return v->isObject() ? &v->toObject() : nullptr;
--- a/deps/spidershim/src/v8context.cc
+++ b/deps/spidershim/src/v8context.cc
@@ -27,7 +27,7 @@
 
 namespace v8 {
 
-Context::Context() : pimpl_(new Impl()) {}
+Context::Context(JSContext* cx) : pimpl_(new Impl(cx)) {}
 
 Context::~Context() { delete pimpl_; }
 
@@ -41,8 +41,7 @@ Local<Context> Context::New(Isolate* isolate,
   if (!cx) {
     return Local<Context>();
   }
-  Context* context = new Context();
-  context->pimpl_->cx = cx;
+  Context* context = new Context(cx);
   JSAutoRequest ar(cx);
   if (!context->CreateGlobal(isolate)) {
     return Local<Context>();
@@ -137,6 +136,28 @@ void Context::Exit() {
   JS_LeaveCompartment(pimpl_->cx, pimpl_->oldCompartment);
   JS_EndRequest(pimpl_->cx);
   GetIsolate()->PopCurrentContext();
+}
+
+void Context::SetEmbedderData(int idx, Local<Value> data)
+{
+	assert(idx >= 0);
+  if (pimpl_->embedderData.length() <= static_cast<unsigned int>(idx) &&
+      !pimpl_->embedderData.resize(idx + 1)) {
+    return;
+  }
+
+  pimpl_->embedderData[idx].set(*GetValue(data));
+}
+
+Local<Value>
+Context::GetEmbedderData(int idx)
+{
+	assert(idx >= 0);
+  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
+    return Local<Value>();
+  }
+
+  return Local<Value>::New(Isolate::GetCurrent(), GetV8Value(pimpl_->embedderData[idx]));
 }
 
 Isolate* Context::GetIsolate() { return Isolate::GetCurrent(); }
--- a/deps/spidershim/src/v8context.cc
+++ b/deps/spidershim/src/v8context.cc
@@ -160,6 +160,30 @@ Context::GetEmbedderData(int idx)
   return Local<Value>::New(Isolate::GetCurrent(), GetV8Value(pimpl_->embedderData[idx]));
 }
 
+void
+Context::SetAlignedPointerInEmbedderData(int idx, void* data)
+{
+  assert((reinterpret_cast<uintptr_t>(data) & 0x1) == 0);
+  assert(idx >= 0);
+  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length() &&
+		  !pimpl_->embedderData.resize(idx + 1)) {
+    return;
+  }
+
+  pimpl_->embedderData[idx].set(JS::PrivateValue(data));
+}
+
+void*
+Context::GetAlignedPointerFromEmbedderData(int idx)
+{
+	assert(idx >= 0);
+  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
+    return nullptr;
+  }
+
+  return pimpl_->embedderData[idx].get().toPrivate();
+}
+
 Isolate* Context::GetIsolate() { return Isolate::GetCurrent(); }
 
 JSContext* JSContextFromContext(Context* context) {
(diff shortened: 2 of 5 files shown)


deps/spidershim/src/conversions.h
```diff
@@ -40,6 +40,10 @@ static inline Value* GetV8Value(JS::Value* val) {
   return reinterpret_cast<Value*>(val);
 }
 
+static inline Value* GetV8Value(JS::MutableHandleValue val) {
```
ehsan, line 43: Hmm, passing in a `MutableHandleValue` here looks wrong, since that type is intended only for out arguments...  As far as I can tell, `GCVector::operator[]` returns a `const JS::Value&' so I'm not sure why you need this overload.

deps/spidershim/src/v8context.cc
```diff
@@ -139,6 +138,52 @@ void Context::Exit() {
...
+
+void*
+Context::GetAlignedPointerFromEmbedderData(int idx)
+{
+	assert(idx >= 0);
+  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
+    return nullptr;
+  }
+
+  return pimpl_->embedderData[idx].get().toPrivate();
```
ehsan, line 184: You need to also check `isPrivate()` above and return null if it fails.

deps/spidershim/src/v8context.cc
```diff
@@ -139,6 +138,52 @@ void Context::Exit() {
...
+    return Local<Value>();
+  }
+
+  return Local<Value>::New(Isolate::GetCurrent(), GetV8Value(pimpl_->embedderData[idx]));
+}
+
+void
+Context::SetAlignedPointerInEmbedderData(int idx, void* data)
+{
+  assert((reinterpret_cast<uintptr_t>(data) & 0x1) == 0);
```
ehsan, line 166: V8 does something to the effect of MOZ_RELEASE_ASSERT here.

Also if I understand their setup correctly I think they enforce 4-byte aligned...  But chakrashim seems to get away without any checks whatsoever!

deps/spidershim/src/conversions.h
tbsaunde, line 43: On Thu, May 12, 2016 at 03:32:20PM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/conversions.h:
> 
> > @@ -40,6 +40,10 @@ static inline Value\* GetV8Value(JS::Value\* val) {
> >    return reinterpret_cast<Value*>(val);
> >  }
> > 
> > +static inline Value\* GetV8Value(JS::MutableHandleValue val) {
> 
> Hmm, passing in a MutableHandleValue here looks wrong, since that type is
> intended only for out arguments... As far as I can tell, GCVector::operator[]
> returns a `const JS::Value&' so I'm not sure why you need this overload.

If you just read GCVector you'd reasonably think that, but haha noap.
Turns out PersistentRooted<T> inherits from PersistentRootedBase<T>
which is specialized for GCVector to inherit from
MutableGCVectorOperations<...>, which defines an operator[] returning
MutableHandle.  So, maybe we could do something else probably involving
some const casting, but just deleting the overload won't work.

deps/spidershim/src/v8context.cc
tbsaunde, line 184: On Thu, May 12, 2016 at 03:41:17PM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8context.cc:
> 
> > -               !pimpl_->embedderData.resize(idx + 1)) {
> > -    return;
> > -  }
> >   +
> > -  pimpl_->embedderData[idx].set(JS::PrivateValue(data));
> >   +}
> >   +
> >   +void*
> >   +Context::GetAlignedPointerFromEmbedderData(int idx)
> >   +{
> > -     assert(idx >= 0);
> > -  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
> > -    return nullptr;
> > -  }
> >   +
> > -  return pimpl_->embedderData[idx].get().toPrivate();
> 
> You need to also check isPrivate() above and return null if it fails.

there is no isPrivate() I think waldo said that is because it uses the
same tag as int32_t.  I'm not sure v8 checks that either.  I'm not sure
this is right, but having examined what node uses this for I can say it
doesn't matter for that use case.

deps/spidershim/src/v8context.cc
tbsaunde, line 166: On Thu, May 12, 2016 at 03:47:04PM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8context.cc:
> 
> > +
> > +Local<Value>
> > +Context::GetEmbedderData(int idx)
> > +{
> > -     assert(idx >= 0);
> > -  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
> > -    return Local<Value>();
> > -  }
> >   +
> > -  return Local<Value>::New(Isolate::GetCurrent(), GetV8Value(pimpl_->embedderData[idx]));
> >   +}
> >   +
> >   +void
> >   +Context::SetAlignedPointerInEmbedderData(int idx, void\* data)
> >   +{
> > -  assert((reinterpret_cast<uintptr_t>(data) & 0x1) == 0);
> 
> V8 does something to the effect of MOZ_RELEASE_ASSERT here.
> 
> Also if I understand their setup correctly I think they enforce 4-byte
> aligned... But chakrashim seems to get away without any checks whatsoever!

v8.h claims 2 byte alignment.  I don't actually see checks in api.cc,
but that's only a problem if the embedder does something they shouldn't
right?

It would be a problem if they used 2 bits, but the docs on Smi layout
suggest they only use 1.

ehsan, line 166: > On Thu, May 12, 2016 at 03:47:04PM -0700, Ehsan Akhgari wrote:
>  In deps/spidershim/src/v8context.cc:
> 
> > +
> > +Local<Value>
> > +Context::GetEmbedderData(int idx)
> > +{
> > -     assert(idx >= 0);
> > -  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
> > -    return Local<Value>();
> > -  }
> >   +
> > -  return Local<Value>::New(Isolate::GetCurrent(), GetV8Value(pimpl_->embedderData[idx]));
> >   +}
> >   +
> >   +void
> >   +Context::SetAlignedPointerInEmbedderData(int idx, void\* data)
> >   +{
> > -  assert((reinterpret_cast<uintptr_t>(data) & 0x1) == 0);
> 
>  V8 does something to the effect of MOZ_RELEASE_ASSERT here.
> 
>  Also if I understand their setup correctly I think they enforce 4-byte
>  aligned... But chakrashim seems to get away without any checks whatsoever!
> v8.h claims 2 byte alignment.  I don't actually see checks in api.cc,
> but that's only a problem if the embedder does something they shouldn't
> right?
> 
> It would be a problem if they used 2 bits, but the docs on Smi layout
> suggest they only use 1.

Yeah I think you're right on both grounds.  Ignore my previous comment.  :-)

deps/spidershim/src/v8context.cc
ehsan, line 184: > On Thu, May 12, 2016 at 03:41:17PM -0700, Ehsan Akhgari wrote:
>  In deps/spidershim/src/v8context.cc:
> 
> > -               !pimpl_->embedderData.resize(idx + 1)) {
> > -    return;
> > -  }
> >   +
> > -  pimpl_->embedderData[idx].set(JS::PrivateValue(data));
> >   +}
> >   +
> >   +void*
> >   +Context::GetAlignedPointerFromEmbedderData(int idx)
> >   +{
> > -     assert(idx >= 0);
> > -  if (static_cast<unsigned int>(idx) >= pimpl_->embedderData.length()) {
> > -    return nullptr;
> > -  }
> >   +
> > -  return pimpl_->embedderData[idx].get().toPrivate();
> 
>  You need to also check isPrivate() above and return null if it fails.
> there is no isPrivate() I think waldo said that is because it uses the
> same tag as int32_t.  I'm not sure v8 checks that either.  I'm not sure
> this is right, but having examined what node uses this for I can say it
> doesn't matter for that use case.

OK, fair enough!

deps/spidershim/src/conversions.h
ehsan, line 43: Ugh.  This is terrible. :(  OK, I guess this hack won't kill us.

Issue #96, opened 2016-05-17
Enable TryCatch tests in TerminateExecution


brendandahl:
This will depend on #52

Issue #95, opened 2016-05-17
Update TerminateExecution tests to use Object/Function Templates


brendandahl:
As a hack we're currently using JSAPI, we should change this to object/function templates when they land.

brendandahl:
Fixed in https://github.com/mozilla/spidernode/commit/eb31b31ce1b4efc8b403bc77f983e57e6eadfac5

Issue #85, opened 2016-05-12
Implement v8::Isolate::TerminateExecution() and v8::Isolate::CancelTerminateExecution()


brendandahl:
part #30

brendandahl:
The way node uses these functions is to execute/eval scripts with a watchdog timeout. The thing that makes this a little tricky is that a node script can recursively run code with timeouts for each new eval context and then resume execution in a parent eval if it catches the timeout exception. 

To properly do this we're going to need to wrap native (function template) calls and make it so they either return true (to continue execution) or false (to stop execution) back into spider monkey based on if execution was terminated, left alone, or termination was cancelled. For now, in my unit tests I'm using JSAPI and manually returning true/false from the native functions, but once function templates are implemented we should change this.

brendandahl:
Fixed in https://github.com/mozilla/spidernode/commit/6e121cc4ece8886aae7bc13185b8926224c5b47a

PR #97, opened 2016-05-17
spidershim: Partially implement isolate terminate execution

brendandahl:
This will need more work when object/function template is implemented.

I filed bugs #95 and #96 and wrote more in https://github.com/mozilla/spidernode/issues/85#issuecomment-219587766

### DIFF
--- a/deps/spidershim/src/v8isolate.h
+++ b/deps/spidershim/src/v8isolate.h
@@ -51,6 +51,8 @@ struct Isolate::Impl {
   mozilla::Maybe<internal::RootStore> eternals;
   std::vector<MessageCallback> messageListeners;
   void* embeddedData[internal::kNumIsolateDataSlots];
+  mozilla::Atomic<bool> serviceInterrupt;
+  mozilla::Atomic<bool> terminatingExecution;
 
   internal::RootStore& EnsurePersistents(Isolate* iso) {
     if (!persistents) {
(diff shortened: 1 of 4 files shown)


deps/spidershim/src/v8isolate.h
```diff
@@ -51,6 +51,8 @@ struct Isolate::Impl {
   mozilla::Maybe<internal::RootStore> eternals;
   std::vector<MessageCallback> messageListeners;
   void* embeddedData[internal::kNumIsolateDataSlots];
+  mozilla::Atomic<bool> serviceInterrupt;
+  mozilla::Atomic<bool> terminatingExecution;
```
ehsan: Out of curiosity, why did you use atomics here?  This is all called on the same thread, right?

deps/spidershim/test/isolate.cc
```diff
@@ -96,3 +97,101 @@ TEST(SpiderShim, IsolateEmbedderData) {
...
+  args.rval().setUndefined();
+  return true;
+}
+
+bool Fail(JSContext *cx, unsigned argc, JS::Value *vp) {
+  EXPECT_TRUE(false);
+  return true;
+}
+
+Local<Value> CompileRun(const char* script) {
```
ehsan, line 115: Can you please use `engine.CompileRun()` instead and take this out?

ehsan:
Looks good!  I submitted a couple of minor nits.

deps/spidershim/src/v8isolate.h
brendandahl: I was following some spider monkey code that does this, thinking that we'd need to support workers/threads in the future. Now looking into it, it seems electron does not support workers.

PR #99, opened 2016-05-17
Enable Address Sanitizer on Travis

ehsan:
I'm going to use this PR to run tests on Travis so that ultimately we can turn on Address Sanitizer on both Linux and OSX.


ehsan:
Fixed in 1532f5d1889581d6ff7a4c02ed382a35e42e75bb.

PR #124, opened 2016-05-20
spidershim: implement v8::Isolate:ThrowException

brendandahl:
I have some tests for this locally (that pass) but they require exposing the private `static Local<T> New(Isolate* isolate, S* that);`, though this won't be needed when we have function/object templates. If wanted I could temporarily make them public, but I think we should just file a follow up issue to enables tests once function/object templates have landed.

### DIFF
--- a/deps/spidershim/src/v8isolate.cc
+++ b/deps/spidershim/src/v8isolate.cc
@@ -219,6 +219,13 @@ void Isolate::CancelTerminateExecution() {
   pimpl_->terminatingExecution = false;
 }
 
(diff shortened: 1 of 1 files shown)


ehsan:
> I have some tests for this locally (that pass) but they require exposing the private static Local<T> New(Isolate\* isolate, S\* that);, though this won't be needed when we have function/object templates. If wanted I could temporarily make them public, but I think we should just file a follow up issue to enables tests once function/object templates have landed.

Sure, sounds good!

PR #128, opened 2016-05-23
Initial implementation of (most of) FunctionTemplate

bzbarsky:
There are still some things missing, some because I'm not sure what their
semantics are, some because they can't really be implemented on top of
SpiderMonkey yet.

Also, we still need a Function::New or some equivalent to make this work...

### DIFF
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -56,6 +56,7 @@ struct JSRuntime;
 struct JSContext;
 class JSScript;
 class V8Engine;
+struct JSClass;
 
 namespace JS {
 class Symbol;
@@ -1822,6 +1823,17 @@ class V8_EXPORT Function : public Object {
   Local<Value> GetName() const;
 
   static Function *Cast(Value *obj);
+
+ protected:
+  static MaybeLocal<Function> New(Local<Context> context,
+                                  FunctionCallback callback,
+                                  Local<Value> data,
+                                  int length,
+                                  Local<FunctionTemplate> templ,
+                                  Local<String> name);
+
+ private:
+  friend class FunctionTemplate;
 };
 
 class V8_EXPORT Promise : public Object {
@@ -2031,7 +2043,8 @@ class V8_EXPORT Template : public Data {
   }
 
  protected:
-  static Local<Template> New(Isolate* isolate);
+  static Local<Template> New(Isolate* isolate,
+                             const JSClass* jsclass = nullptr);
 
  private:
   // We treat Templates as JS::Values containing a JSObject* that we can
@@ -2064,6 +2077,14 @@ class V8_EXPORT FunctionTemplate : public Template {
                       Handle<Value> data = Handle<Value>());
   bool HasInstance(Handle<Value> object);
   void Inherit(Handle<FunctionTemplate> parent);
+
+  // XXXbz This is public because we call it from inside some static functions
+  // over in Function.  Maybe those should become Function members and then we
+  // can just friend Function and make this protected.
+  Local<Object> CreateNewInstance();
+
+ private:
+  Local<ObjectTemplate> FetchOrCreateTemplate(size_t slotIndex);
 };
 
 enum class PropertyHandlerFlags {
@@ -2183,6 +2204,7 @@ class V8_EXPORT ObjectTemplate : public Template {
   friend struct FunctionCallbackData;
   friend struct FunctionTemplateData;
   friend class Utils;
+  friend class FunctionTemplate;
 
   Local<Object> NewInstance(Handle<Object> prototype);
   Handle<String> GetClassName();
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -2200,6 +2200,7 @@ class V8_EXPORT ObjectTemplate : public Template {
   void SetCallAsFunctionHandler(FunctionCallback callback,
                                 Handle<Value> data = Handle<Value>());
 
+  class InstanceClass;
  private:
   friend struct FunctionCallbackData;
   friend struct FunctionTemplateData;
@@ -2208,6 +2209,11 @@ class V8_EXPORT ObjectTemplate : public Template {
 
   Local<Object> NewInstance(Handle<Object> prototype);
   Handle<String> GetClassName();
+
+  // GetInstanceClass creates it if needed, otherwise returns the existing
+  // one.  Null is returned if objects should be created with the default
+  // plain-object JSClass.
+  InstanceClass* GetInstanceClass();
 };
 
 class V8_EXPORT External : public Value {
--- a/deps/spidershim/src/conversions.h
+++ b/deps/spidershim/src/conversions.h
@@ -83,4 +83,36 @@ static inline JSString* GetString(const Value* val) {
 static inline JSString* GetString(const Local<Value>& val) {
   return GetString(*val);
 }
+
+// Various callbacks cannot be represented in a single JS::Value, because the
+// max size of the stored private is 63 bits and code addresses can have their
+// least significant bit set, whereas JS::PrivateValue assumes the thing being
+// stored has the LSB clear.  So we need to encode FunctionCallback as two
+// private values.  The first one stores all but the least significant bit,
+// while the second one stores the LSB left-shifted by one.
+template<typename T>
+static inline T ValuesToCallback(JS::HandleValue val1,
+                                 JS::HandleValue val2) {
+  void* ptr1 = val1.toPrivate();
+  void* ptr2 = val2.toPrivate();
+  assert(ptr2 == reinterpret_cast<void*>(0x0) ||
+         ptr2 == reinterpret_cast<void*>(0x1 << 1));
+  return reinterpret_cast<T>
+    (reinterpret_cast<uintptr_t>(ptr1) |
+     (reinterpret_cast<uintptr_t>(ptr2) >> 1));
+}
+
+template<typename T>
+static inline void CallbackToValues(T callback,
+                                    JS::MutableHandleValue val1,
+                                    JS::MutableHandleValue val2) {
+  void* ptr1 = reinterpret_cast<void*>
+    (reinterpret_cast<uintptr_t>(callback) & ~uintptr_t(0x1));
+  void* ptr2 = reinterpret_cast<void*>
+    (((reinterpret_cast<uintptr_t>(callback) & uintptr_t(0x1)) << 1));
+  val1.setPrivate(ptr1);
+  val2.setPrivate(ptr2);
+  assert(ValuesToCallback<T>(val1, val2) == callback);
+}
+
 }
--- a/deps/spidershim/src/conversions.h
+++ b/deps/spidershim/src/conversions.h
@@ -61,6 +61,10 @@ static inline JSObject* GetObject(Value* val) {
   return v->isObject() ? &v->toObject() : nullptr;
 }
 
+static inline JSObject* GetObject(Template* val) {
+  return GetObject(GetV8Value(val));
+}
+
 static inline JSObject* GetObject(const Value* val) {
   const JS::Value* v = GetValue(val);
   return v->isObject() ? &v->toObject() : nullptr;
--- a/deps/spidershim/src/v8function.cc
+++ b/deps/spidershim/src/v8function.cc
@@ -112,32 +112,67 @@ bool NativeFunctionCallback(JSContext* cx, unsigned argc, JS::Value* vp) {
   calleeVal.setObject(args.callee());
   Local<Function> calleeFunction =
     internal::Local<Function>::New(isolate, calleeVal);
-  Local<Object> _this =
-    internal::Local<Object>::New(isolate, args.thisv());
   Local<Value> data = GetHiddenCalleeData(cx, callee);
-  mozilla::UniquePtr<Value*[]> v8args(new Value*[argc]);
-  for (unsigned i = 0; i < argc; ++i) {
-    Local<Value> arg = internal::Local<Value>::New(isolate, args[i]);
-    if (arg.IsEmpty()) {
-      return false;
+  JS::RootedValue templateVal(cx, js::GetFunctionNativeReserved(callee, 0));
+  Local<FunctionTemplate> templ;
+  if (!templateVal.isUndefined()) {
+    templ = internal::Local<FunctionTemplate>::NewTemplate(isolate,
+                                                           templateVal);
+  }
+
+  Local<Object> _this;
+  if (args.isConstructing()) {
+    if (templ.IsEmpty()) {
+      JS::RootedObject newObj(cx, JS_NewPlainObject(cx));
+      if (!newObj) {
+        return false;
+      }
+      _this = internal::Local<Object>::New(isolate, JS::ObjectValue(*newObj));
+    } else {
+      _this = templ->CreateNewInstance();
+      if (_this.IsEmpty()) {
+        return false;
+      }
     }
-    v8args[i] = *arg;
-  }
-  // TODO: Figure out what we want to do for holder.  See
-  // https://groups.google.com/d/msg/v8-users/Axf4hF_RfZo/hA6Mvo78AqAJ
-  FunctionCallbackInfo<Value> info(v8args.get(), argc, _this, _this,
-                                   args.isConstructing(),
-                                   data, calleeFunction);
-  FunctionCallback callback = GetHiddenCallback(cx, callee);
-  if (!callback) {
-    return false;
-  }
-  callback(info);
-  if (auto rval = info.GetReturnValue().Get()) {
-    args.rval().set(*GetValue(rval));
   } else {
-    args.rval().setUndefined();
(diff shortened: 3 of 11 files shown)


deps/spidershim/include/v8.h
```diff
@@ -2037,7 +2037,9 @@ class V8_EXPORT Template : public Data {
   }
 
  protected:
-  static Local<Template> New(Isolate* isolate);
+  // jsclass if a void* so we can avoid having js::Class around here.
+  static Local<Template> New(Isolate* isolate,
+			     const void* jsclass = nullptr);
```
ehsan: Forward-declaring `js::Class` here should be fine.

deps/spidershim/src/conversions.h
```diff
@@ -61,6 +61,11 @@ static inline JSObject* GetObject(Value* val) {
   return v->isObject() ? &v->toObject() : nullptr;
 }
 
+static inline JSObject* GetObject(Template* val) {
+  // XXXbz Can we assume val is never null here?  Seems dubious.
```
ehsan: Most of these helpers assume their argument is non-null, and leave it up to the caller to ensure that.  I'd be fine with assuming the same here.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+
+// static
+Local<FunctionTemplate> FunctionTemplate::New(Isolate* isolate,
+                                              FunctionCallback callback,
+                                              Handle<Value> data,
+                                              Handle<Signature> signature,
+                                              int length)
+{
+  Local<Template> templ = Template::New(isolate, &functionTemplateClass);
+  if (*templ) {
```
ehsan: Hmm, this looks wrong.  I think you meant to check `templ.IsEmpty()`?

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+    return Local<FunctionTemplate>();
+  }
+
+  JSContext* cx = JSContextFromIsolate(isolate);
+  JS::RootedObject obj(cx, GetObject(*templ));
```
ehsan: FWIW this root here is not needed since anything wrapped in a `Local<>` is already rooted, and `SetCallbackAndData` doesn't require a `Handle`.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+  assert(obj);
+
+  SetCallbackAndData(obj, callback, data);
+
+  // TODO: figure out what should happen with signature.  What is it supposed to
+  // do?
```
ehsan: The `ReceiverSignature` V8 test might help here.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+}
+
+MaybeLocal<Function> FunctionTemplate::GetFunction(Local<Context> context)
+{
+  JSContext* cx = JSContextFromContext(*context);
+  JS::RootedObject obj(cx, GetObject(this));
+  assert(obj);
+
+  // XXXbz V8 uses a temporary HandleScope here, but I _think_ that's some sort
+  // of performance optimization...
```
ehsan, line 118: Yeah, should be!

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+  JS::RootedValue callbackVal(cx, js::GetReservedSlot(obj, CallbackSlot));
+  // XXXbz the .get() here is working around
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1275082
+  auto callback = reinterpret_cast<FunctionCallback>(callbackVal.get().toPrivate());
+
+  JS::RootedValue dataVal(cx, js::GetReservedSlot(obj, DataSlot));
+  // XXXbz is this a sane thing to do?  Or do I need to register this thing with
+  // some HandleScope?  Seems like I need to do the latter if the callee has
+  // this thing outlive the Rooted, right?
+  Local<Value> dataV8Val = GetV8Value(dataVal.address());
```
ehsan: The `conversions.h` routines can only be used for stuff currently registered with a `HandleScope`.  What you want to do here instead is something like:

```
Local<Value> dataV8Val = internal::Local<Value>::New(isolate, js::getReservedSlot(...));
```

This way you won't need the `dataVal` root either.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+  JS::RootedValue lengthVal(cx, js::GetReservedSlot(obj, LengthSlot));
+
+  // Figure out our prototype's prototype object.
+  JS::RootedObject protoProto(cx);
+  JS::RootedValue parentVal(cx, js::GetReservedSlot(obj, ParentSlot));
+  if (!parentVal.isUndefined()) {
+    assert(parentVal.isObject());
+    // XXXbz this seems a bit fishy... is it?
+    auto parentTemplate =
+      static_cast<FunctionTemplate*>(GetV8Template(parentVal.address()));
```
ehsan: You should do something like the above here too.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+    MaybeLocal<Function> parentFunc = parentTemplate->GetFunction(context);
+    if (parentFunc.IsEmpty()) {
+      return MaybeLocal<Function>();
+    }
+
+    JS::RootedObject parentObj(cx, GetObject(parentFunc.ToLocalChecked()));
+    assert(parentObj);
+
+    JS::RootedValue protoProtoVal(cx);
+    if (!JS_GetProperty(cx, parentObj, "prototype", &protoProtoVal)) {
```
ehsan, line 154: I'm wondering whether you should use `JS_GetPrototype()` here...

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+      // XXXbz what to do with the exception?
```
ehsan: We usually don't do anything in case of errors like this...  This probably needs fixing overall at some point.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+  JS::RootedValue protoTemplateVal(cx, js::GetReservedSlot(obj, ProtoSlot));
+  if (protoTemplateVal.isUndefined()) {
+    protoObj = JS_NewObjectWithGivenProto(cx, nullptr, protoProto);
+    if (!protoObj) {
+      // XXXbz what to do with the exception?
+      return MaybeLocal<Function>();
+    }
+  } else {
+    assert(protoTemplateVal.isObject());
+    // XXXbz this seems a bit fishy... is it?
```
ehsan: It is.  :-)

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+      // XXXbz what to do with the exception?
+      return MaybeLocal<Function>();
+    }
+  }
+
+  // TODO: We should be doing something faster here in terms of setting up
+  // prototypes right from the beginning or something.  Maybe we can refactor
+  // Function::New's implementation a bit to allow us to pass the right things
+  // into there.
+  Local<Function> func = Function::New(context->GetIsolate(),
```
ehsan: FWIW `Function::New` looks like the right thing to use here.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+  JS::RootedObject funcObj(cx, GetObject(func));
+  assert(funcObj);
+  if (!JS_CopyPropertiesFrom(cx, funcObj, obj)) {
+    // XXXbz What do we do with the exception?
+    return MaybeLocal<Function>();
+  }
+
+  // TODO: Is the result of all this supposed to have a non-configurable
+  // .prototype like builtins, or a configurable one like scripted functions?
+  // Let's assume the former.
```
ehsan: Please file follow-up issues for investigations such as this.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+  JSContext* cx = JSContextFromIsolate(isolate);
+  JS::RootedObject obj(cx, GetObject(this));
+  assert(obj);
+
+  JS::RootedValue protoTemplateVal(cx, js::GetReservedSlot(obj, ProtoSlot));
+  if (protoTemplateVal.isObject()) {
+    // XXX bz is this the right way to do this?
+    Local<Template> templ =
+      internal::Local<Template>::NewTemplate(isolate, protoTemplateVal);
+    return Local<ObjectTemplate>(static_cast<ObjectTemplate*>(*templ));
```
ehsan: No, try this instead:

```
return internal::Local<ObjectTemplate>::NewTemplate(isolate, protoTemplateVal);
```

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+    return Local<ObjectTemplate>();
+  }
+
+  js::SetReservedSlot(obj, ProtoSlot, JS::ObjectValue(*GetObject(*templ)));
+  return templ;
+}
+
+// TODO: Need to implement SetClassName.  Does this map to JSClass names, and if
+// so do we need to maintain our own (heap-allocated, presumably) instance
+// JSClass?
```
ehsan: The answer to the first question at least seems to be yes.  See `TestFunctionTemplateAccessor` for example where it sets the `"funky"` class name...

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+
+// TODO: Need to implement SetHiddenPrototype.  How?  SpiderMonkey doesn't
+// really have such a concept.
```
ehsan: As discussed before, I think we can leave this one.

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+void
+FunctionTemplate::SetCallHandler(FunctionCallback callback,
+                                 Handle<Value> data)
+{
+  JSContext* cx = JSContextFromIsolate(Isolate::GetCurrent());
+  JS::RootedObject obj(cx, GetObject(this));
+  assert(obj);
+
+  if (TemplateIsInstantiated(obj)) {
+    // TODO: Supposed to be a fatal error of some sort.
```
ehsan, line 262: Like a MOZ_CRASH?

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -0,0 +1,315 @@
...
+
+  SetCallbackAndData(obj, callback, data);
+}
+
+// TODO: Need to implement HasInstance.  What are the actual semantics of this?
+// I can't figure out so far from the V8 API docs and their implementation.  In
+// particular, does it just check whether this is precisely an object that
+// constructing via the function returned by this FunctionTemplate produces?  Or
+// is any subclass (including via FunctionTemplates that inherit from this one)
+// OK?
```
ehsan: `InstanceCheckOnInstanceAccessor` might help here.

ehsan:
Please see my comments above!  Given that this setup is fairly involved, it would also be great if you'd write some docs.  Thanks!

deps/spidershim/src/v8functiontemplate.cc
bzbarsky, line 154: No, definitely not.  I'm getting function.prototype, not function.**proto**.  This is the object that's supposed to be used as the proto for instance objects created when using this function as a ctor.

deps/spidershim/src/v8functiontemplate.cc
bzbarsky: https://github.com/mozilla/spidernode/issues/140

deps/spidershim/src/v8functiontemplate.cc
bzbarsky, line 262: No, V8 seems to have some way of catching fatal errors like this, from reading their code...  ApiCheck is the relevant thing.

bzbarsky:
OK, at this point I think this is landable with other things as followups.  One big question: given a `FunctionTemplate` `tmpl`, should `tmpl->InstanceTemplate()->NewInstance()` be equivalent to `tmpl->GetFunction()->NewInstance()`?  That is, should the `InstanceTemplate()` know it's the instance template of a `FunctionTemplate` and act accordingly when someone tries to instantiate it?

The V8 API tests on V8 tip think so.  Note some of the commented-out bits in test/v8function.cc.  I didn't implement that behavior so far, but it would actually be quite easy to do if we decide we want it.  We'd just need a backpointer from the `ObjectTemplate` to the `FunctionTemplate` it's an instance template for.

Note that chakrashim does NOT have this behavior as far as I can tell, which is why I didn't do it for now.  This behavior may have been a V8 API change....

deps/spidershim/include/v8.h
```diff
@@ -2064,6 +2077,14 @@ class V8_EXPORT FunctionTemplate : public Template {
                       Handle<Value> data = Handle<Value>());
   bool HasInstance(Handle<Value> object);
   void Inherit(Handle<FunctionTemplate> parent);
+
+  // XXXbz This is public because we call it from inside some static functions
+  // over in Function.  Maybe those should become Function members and then we
+  // can just friend Function and make this protected.
```
ehsan, line 2083: Seems like this only gets called from `NativeFunctionCallback`.  I filed #145 to fix this.

deps/spidershim/src/v8function.cc
```diff
@@ -242,17 +264,46 @@ MaybeLocal<Function> Function::New(Local<Context> context,
...
+                                   FunctionCallback callback,
+                                   Local<Value> data,
+                                   int length,
+                                   Local<FunctionTemplate> templ,
+                                   Local<String> name) {
   JSContext* cx = JSContextFromContext(*context);
-  JSFunction* func = JS_NewFunction(cx, NativeFunctionCallback, length, 0, nullptr);
+  JSFunction* func;
+  // It's a bit weird to always pass JSFUN_CONSTRUCTOR, but it's not clear to me
+  // when we would want things we get from her to NOT be constructible...  Maybe
```
ehsan: Nit: *here.

deps/spidershim/src/v8function.cc
```diff
@@ -242,17 +264,46 @@ MaybeLocal<Function> Function::New(Local<Context> context,
...
+  // when templ.IsEmpty()?  Hard to tell what chackrashim does because they
+  // don't seem to implement Function::New at all and all their FunctionTemplate
+  // stuff seems constructible at first glance.
```
ehsan: Hmm, yeah this seems like the right thing to do.

deps/spidershim/include/v8.h
bzbarsky, line 2083: Correct.  It's not hard to fix; I'm happy to just do it in a followup.

deps/spidershim/src/v8function.cc
bzbarsky: Fixed.

deps/spidershim/src/v8objecttemplate.cc
```diff
@@ -22,11 +22,430 @@
...
+bool NativeAccessorCallback(JSContext* cx, unsigned argc, JS::Value* vp) {
+  using namespace v8;
+  Isolate* isolate = Isolate::GetCurrent();
+  JS::CallArgs args = JS::CallArgsFromVp(argc, vp);
+  JS::RootedObject callee(cx, &args.callee());
+
+  // XXXbz This seems busted conceptually: "thisv" can be a non-object!  But
+  // objects are all PropertyCallbackInfo accepts....
+  Local<Object> thisObject =
+    internal::Local<Object>::New(isolate, args.thisv());
```
ehsan, line 189: Both here and in `NativeFunctionCallback`, we should check to make sure `thisv` is actually an object.

deps/spidershim/src/v8objecttemplate.cc
```diff
@@ -22,11 +22,430 @@
...
+  if (prototype.IsEmpty()) {
+    JS::RootedObject currentGlobal(cx, JS::CurrentGlobalOrNull(cx));
+    protoObj = JS_GetObjectPrototype(cx, currentGlobal);
+  } else {
+    protoObj = &GetValue(prototype)->toObject();
+  }
+
+  InstanceClass* instanceClass = GetInstanceClass();
+
+  // XXXbz this needs more fleshing out to deal with the whole business of
```
ehsan, line 303: Do you mind clarifying this comment a bit, please?

bzbarsky, line 303: Done.

deps/spidershim/src/v8objecttemplate.cc
bzbarsky, line 189: Done.

PR #137, opened 2016-05-27
Proxy

tbsaunde:

### DIFF
--- a/deps/spidershim/src/v8proxy.cc
+++ b/deps/spidershim/src/v8proxy.cc
@@ -35,4 +35,14 @@ Proxy::Cast(Value* v)
   assert(v->IsProxy());
   return static_cast<Proxy*>(v);
 }
+
+Local<Object>
+Proxy::GetTarget()
+{
+  JSContext* cx = JSContextFromIsolate(Isolate::GetCurrent());
+  JS::RootedObject obj(cx, GetObject(this));
+  JS::Value target;
+  target.setObject(*GetProxyTargetObject(obj));
+  return internal::Local<Object>::New(Isolate::GetCurrent(), target);
+}
 }
--- a/deps/spidershim/src/v8proxy.cc
+++ b/deps/spidershim/src/v8proxy.cc
@@ -1,0 +1,38 @@
+// Copyright Mozilla Foundation. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


deps/spidershim/src/v8proxy.cc
```diff
@@ -0,0 +1,55 @@
...
+{
+  assert(v->IsProxy());
+  return static_cast<Proxy*>(v);
+}
+
+Local<Object>
+Proxy::GetTarget()
+{
+  JSContext* cx = JSContextFromIsolate(Isolate::GetCurrent());
+  JS::RootedObject obj(cx, GetObject(this));
```
ehsan, line 43: This root is unnecessary, you can just pass the `JSObject*` to `GetProxyTargetObject` directly.

deps/spidershim/src/v8proxy.cc
```diff
@@ -0,0 +1,55 @@
...
+  target.setObject(*GetProxyTargetObject(obj));
+  return internal::Local<Object>::New(Isolate::GetCurrent(), target);
+}
+
+Local<Value>
+Proxy::GetHandler()
+{
+  // XXX implement this when spider monkey exposes it.
+  return Local<Value>();
+}
```
ehsan, line 54: Looks like Node actually uses this.  Let's leave this method out so that we remember to add the full implementation when SpiderMonkey support has been added.  What's the bug number for SpiderMonkey side of things?

deps/spidershim/src/v8value.cc
```diff
@@ -324,4 +325,15 @@ bool Value::IsNativeError() const {
 }
 
 bool Value::IsExternal() const { return External::IsExternal(this); }
+
+bool
+Value::IsProxy() const
+{
+  JSObject* obj = GetObject(this);
```
ehsan, line 332: You need to check `IsObject()` first.

ehsan:
Can you please add some tests too?  Thanks!

deps/spidershim/src/v8value.cc
tbsaunde, line 332: On Mon, May 30, 2016 at 06:11:18AM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8value.cc:
> 
> > @@ -324,4 +325,15 @@ bool Value::IsNativeError() const {
> >  }
> > 
> >  bool Value::IsExternal() const { return External::IsExternal(this); }
> > +
> > +bool
> > +Value::IsProxy() const
> > +{
> > -  JSObject\* obj = GetObject(this);
> 
> You need to check IsObject() first.

GetObject() checks it

ehsan, line 332: Oh, right!  Nevermind then.

deps/spidershim/src/v8proxy.cc
tbsaunde, line 54: On Mon, May 30, 2016 at 06:10:48AM -0700, Ehsan Akhgari wrote:

> In deps/spidershim/src/v8proxy.cc:
> 
> > +Local<Object>
> > +Proxy::GetTarget()
> > +{
> > -  JSContext\* cx = JSContextFromIsolate(Isolate::GetCurrent());
> > -  JS::RootedObject obj(cx, GetObject(this));
> > -  JS::Value target;
> > -  target.setObject(*GetProxyTargetObject(obj));
> > -  return internal::Local<Object>::New(Isolate::GetCurrent(), target);
> >   +}
> >   +
> >   +Local<Value>
> >   +Proxy::GetHandler()
> >   +{
> > -  // XXX implement this when spider monkey exposes it.
> > -  return Local<Value>();
> >   +}
> 
> Looks like Node actually uses this. Let's leave this method out so that we
> remember to add the full implementation when SpiderMonkey support has been

 As far as I can tell it only uses it for a debugging tool thing, so it
 doesn't really seem important.

> added. What's the bug number for SpiderMonkey side of things?

apparently they had patches lying around but forgot to file and I was
bad at pinging so bug 1277949.

ehsan:
Ping?

tbsaunde:
pushed with the rooting thing fixed, we can deal with the rest of it when we update sm.

PR #160, opened 2016-06-09
spidershim: Fix FunctionTemplate::HasInstance

### LINKED ISSUE #152: HasInstance implementation doesn't seem right
I'm looking at https://github.com/mozilla/spidernode/commit/995c667c03925785024ff1b066bf05747bad204b and I see two problems, I think:

1)  There is no reason whatsoever that we can trust what `JS_GetProperty(cx, protoObj, "constructor", &funcVal)` returns on the object.  It can be absolutely anything if proto chains got messed with.  So if we're relying on this check to tell us something about the internal slots of the value involved (c.f. the desire to use this stuff in Signature over in https://github.com/mozilla/spidernode/issues/144), this is not the way to go.

2)  Related to the above, since there is no guarantee of us getting a function we know anything about from `.constructor` calling `GetFunctionNativeReserved` on it is not OK.  It might not even have those extra reserved slots!

@ehsan am I missing something here?

ehsan:
This should fix #152.

### DIFF
--- a/deps/spidershim/src/v8functiontemplate.cc
+++ b/deps/spidershim/src/v8functiontemplate.cc
@@ -417,16 +417,18 @@ bool FunctionTemplate::HasInstance(Local<Value> val) {
   Isolate* isolate = Isolate::GetCurrent();
   JSContext* cx = JSContextFromIsolate(isolate);
   AutoJSAPI jsAPI(cx, this);
-  Local<Object> obj = val.As<Object>();
+  JS::RootedValue value(cx, *GetValue(val));
+  if (!JS_WrapValue(cx, &value)) {
+    return false;
+  }
+  Local<Object> obj = internal::Local<Object>::New(isolate, value);
   if (!ObjectTemplate::IsObjectFromTemplate(obj)) {
     return false;
   }
-  JS::RootedValue thisVal(cx, *GetValue(this));
   Local<FunctionTemplate> ctor = ObjectTemplate::GetObjectTemplateConstructor(obj);
   while (!ctor.IsEmpty()) {
-    JS::RootedValue ctorVal(cx, *GetValue(*ctor));
-    bool equals = false;
-    if (JS_StrictlyEqual(cx, ctorVal, thisVal, &equals) && equals) {
+    // We don't need JS_StrictlyEqual since we know both sides are objects here.
+    if (*GetValue(ctor) == *GetValue(this)) {
       return true;
     }
     ctor = GetParent().FromMaybe(Local<FunctionTemplate>());
@@ -440,7 +442,7 @@ Local<Value> FunctionTemplate::MaybeConvertObjectProperty(Local<Value> value,
   if (!value.IsEmpty() && value->IsObject() &&
       JS_GetClass(GetObject(value)) == &functionTemplateClass) {
     Local<Function> func =
-      reinterpret_cast<FunctionTemplate*>(GetValue(*value))->GetFunction();
+      reinterpret_cast<FunctionTemplate*>(GetValue(value))->GetFunction();
     func->SetName(name);
     return func;
   }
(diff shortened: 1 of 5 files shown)


deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -417,31 +417,19 @@ bool FunctionTemplate::HasInstance(Local<Value> val) {
...
+  if (!ObjectTemplate::IsObjectFromTemplate(obj)) {
     return false;
   }
   JS::RootedValue thisVal(cx, *GetValue(this));
-  JS::RootedValue templateVal(cx,
-    js::GetFunctionNativeReserved(&funcVal.toObject(), 0));
-  while (templateVal.isObject()) {
+  Local<FunctionTemplate> ctor = ObjectTemplate::GetObjectTemplateConstructor(obj);
+  while (!ctor.IsEmpty()) {
+    JS::RootedValue ctorVal(cx, *GetValue(*ctor));
```
bzbarsky: I'm getting pretty tired of having to write GetValue(*someTemplate).  Want to add some more overloads to conversions.h?  ;)

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -417,31 +417,19 @@ bool FunctionTemplate::HasInstance(Local<Value> val) {
...
     bool equals = false;
-    if (JS_StrictlyEqual(cx, templateVal, thisVal, &equals) && equals) {
-      Local<ObjectTemplate> instanceTemplate = InstanceTemplate();
-      if (instanceTemplate.IsEmpty()) {
-        return false;
-      }
-      return instanceTemplate->IsInstance(obj);
+    if (JS_StrictlyEqual(cx, ctorVal, thisVal, &equals) && equals) {
```
bzbarsky: Honestly, I think you can just test `ctorVal == thisVal`....  The cases in which that doesn't do the same thing as JS_StrictlyEqual all involve primitives (`-0`, `NaN`, two JSStrings that are data-equal but not pointer-equal), but we know we're dealing with objects, so just comparing the pointer values is enough, and Value::operator== will do that fine.  Then you don't have to worry about the (unreachable, in this case) situation in which JS_StrictlyEqual returns false.

deps/spidershim/src/v8objecttemplate.cc
```diff
@@ -383,4 +386,22 @@ Local<FunctionTemplate> ObjectTemplate::GetConstructor() {
   return internal::Local<FunctionTemplate>::NewTemplate(isolate, ctorVal);
 }
 
+bool ObjectTemplate::IsObjectFromTemplate(Local<Object> object) {
+  return !!(JS_GetClass(GetObject(object))->flags & InstanceClass::instantiatedFromTemplate);
```
bzbarsky, line 391: Add an assert about !object.IsEmpty(), perhaps?

deps/spidershim/src/v8objecttemplate.cc
```diff
@@ -383,4 +386,22 @@ Local<FunctionTemplate> ObjectTemplate::GetConstructor() {
...
+}
+
+Local<FunctionTemplate> ObjectTemplate::GetObjectTemplateConstructor(Local<Object> object) {
```
bzbarsky, line 394: Document in the header that the arg should have tested true for IsObjectFromTemplate?

deps/spidershim/src/v8functiontemplate.cc
```diff
@@ -417,31 +417,19 @@ bool FunctionTemplate::HasInstance(Local<Value> val) {
   Isolate* isolate = Isolate::GetCurrent();
   JSContext* cx = JSContextFromIsolate(isolate);
   AutoJSAPI jsAPI(cx, this);
```
bzbarsky, line 419: You should probably ensure that obj is wrapped into our compartment at this point or something.  Otherwise we can have situations where we're coming in on some compartment different from the one of `this` and the arg is actually a cross-compartment wrapper for something that _is_ an instance of this template... but a CCW will test false for IsObjectFromTemplate.

bzbarsky:
Looks good.

ehsan:
Merged in ef8abcc.

PR #165, opened 2016-07-06
Implement signature checking for functions and properties

### LINKED ISSUE #144: Figure out what to do with Signature in FunctionTemplate stuff
It's supposed to enforce some constraint on ... well what?  Sounds kind of like on the holder, which is a fairly useless thing in practice.

Anyway, looking at chakrashim, they set the "holder" for their calls to the "this" value unless there is a Signature.  If there is, they call Utils::CheckSignature, which always sets "holder" to "this", but ensure that the object involved is an instance of the instance template of the function template that was originally passed to Signature::New (which only accepts conditions on "this", not on args, in chakrashim).  They do this by storing the object template used to construct each object in that object, and just checking that the stored thing is the instance template involved.

We could do something similar: store the Signature (aka FunctionTemplate) in our JSFunction, then if there is one get its instance template's JSClass and compare to it.  This would require us to always use a custom JSClass for instances, of course.

ehsan:
This fixes #144.  For reviewing, it probably helps to read the individual commits in order.  The last two are the logical equivalent of the second and third commits, but for properties instead of functions.

@bzbarsky can you please review?  Thanks!

### DIFF
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -2322,8 +2322,7 @@ class V8_EXPORT Signature : public Data {
  public:
   static Local<Signature> New(
       Isolate* isolate,
-      Handle<FunctionTemplate> receiver = Handle<FunctionTemplate>(),
-      int argc = 0, Handle<FunctionTemplate> argv[] = nullptr);
+      Handle<FunctionTemplate> receiver = Handle<FunctionTemplate>());
 
  private:
   Signature();
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -2175,6 +2175,9 @@ class V8_EXPORT FunctionTemplate : public Template {
   // InstanceTemplates of all ancestors, on the given object.  Returns false on
   // failure, true on success.
   bool InstallInstanceProperties(Local<Object> target);
+  // Performs a signature check if needed.  Returns true and a holder object
+  // unless the signature check fails.
+  bool CheckSignature(Local<Object> _this, Local<Object>& holder);
 };
 
 enum class PropertyHandlerFlags {
--- a/deps/spidershim/include/v8.h
+++ b/deps/spidershim/include/v8.h
@@ -770,6 +770,7 @@ class V8_EXPORT HandleScope {
 
   static Value* AddToScope(Value* val);
   static Template* AddToScope(Template* val);
+  static Signature* AddToScope(Signature* val);
   static Script* AddToScope(JSScript* script, Local<Context> context);
   static Private* AddToScope(JS::Symbol* priv);
   static Message* AddToScope(Message* msg);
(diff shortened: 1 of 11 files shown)


deps/spidershim/src/accessor.h
```diff
@@ -101,6 +107,30 @@ struct CallbackTraits<typename XerTraits<N>::Setter, N> {
...
 
+struct SignatureChecker {
+  // Performs a signature check if needed.  Returns true and a holder object
+  // unless the signature check fails.
+  static bool CheckSignature(JS::HandleObject accessorData,
+                             v8::Local<Object> _this,
+                             v8::Local<Object>& holder) {
+    Isolate* isolate = Isolate::GetCurrent();
+    JSContext* cx = JSContextFromIsolate(isolate);
+    AutoJSAPI jsAPI(cx, GetObject(_this));
```
bzbarsky, line 118: Are we always guaranteed that `_this` and `accessorData` are same-compartment?  I guess we are because the `accessorData` is in the compartment of the accessor function by construction and `_this` is in that same compartment, ok...

deps/spidershim/src/accessor.h
```diff
@@ -211,13 +253,13 @@ bool SetAccessor(JSContext* cx,
   // TODO: What should happen with "settings", "signature"?  See
```
bzbarsky, line 253: Nix "signature" from this comment.

deps/spidershim/test/template.cc
```diff
@@ -441,6 +438,7 @@ static void TestSignature(V8Engine& engine, Local<Context> context,
   }
 }
 
+#if 0
```
bzbarsky, line 441: This still leaves the actual test #if 0, no?  Why?

deps/spidershim/include/v8.h
```diff
@@ -2175,6 +2175,9 @@ class V8_EXPORT FunctionTemplate : public Template {
   // InstanceTemplates of all ancestors, on the given object.  Returns false on
   // failure, true on success.
   bool InstallInstanceProperties(Local<Object> target);
+  // Performs a signature check if needed.  Returns true and a holder object
+  // unless the signature check fails.
+  bool CheckSignature(Local<Object> _this, Local<Object>& holder);
```
bzbarsky, line 2182: It might be good to name the first argument thisObj or something, and clearly document what it represents.

deps/spidershim/test/template.cc
ehsan, line 441: Yes.  That test also requires `SetAccessorProperty` before it can pass, which I'm implementing on top of this...  I will eventually remove the `#if 0` once that patch is ready.

deps/spidershim/src/accessor.h
ehsan, line 118: Yes.

deps/spidershim/src/accessor.h
ehsan, line 253: Will do.

deps/spidershim/include/v8.h
ehsan, line 2182: Will do.

ehsan:
Merged manually: https://github.com/mozilla/spidernode/compare/2f0d4c2efb1c...8a1f864

PR #167, opened 2016-07-12
Implement {Object,Template}::SetAccessorProperty

ehsan:
These commits are mostly straightforward.  The most interesting part is the second commit, which fixes our signature checking to support checking an object created from `FunctionTemplate` B inherited from A when the signature check is performed on A.  This case is examined by the `sub_fun_instance` test inside `FunctionTemplateReceiverSignature`.

@bzbarsky can you please review?  Thanks!

### DIFF
--- a/deps/spidershim/src/accessor.h
+++ b/deps/spidershim/src/accessor.h
@@ -171,8 +171,8 @@ bool NativeAccessorCallback(JSContext* cx, unsigned argc, JS::Value* vp) {
     v8::Local<Object> holder;
     if (!SignatureChecker::CheckSignature(accessorData, thisObject, holder)) {
       isolate->ThrowException(
-          Exception::Error(String::NewFromUtf8(isolate,
-                                               kIllegalInvocation)));
+          Exception::TypeError(String::NewFromUtf8(isolate,
+                                                   kIllegalInvocation)));
       return false;
     }
     typedef typename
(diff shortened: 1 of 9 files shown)


deps/spidershim/src/accessor.h
```diff
@@ -254,35 +284,20 @@ bool SetAccessor(JSContext* cx,
   // TODO: What should happen with "settings"?  See
   // https://github.com/mozilla/spidernode/issues/141
 
-  JS::RootedObject getterObj(cx, CreateAccessor(cx, getter, name, data, signature));
+  JSObject* getterObj = CreateAccessor(cx, getter, name, data, signature);
```
bzbarsky, line 287: No, this change is wrong.  getterObj is now unrooted across the call to CreateAccessor for the setter, which can GC.  This needs to stay rooted.

deps/spidershim/src/accessor.h
```diff
@@ -240,6 +240,36 @@ JSObject* CreateAccessor(JSContext* cx, CallbackType callback,
   return funObj;
 }
 
+template<class N>
+bool SetAccessor(JSContext* cx,
+                 JS::HandleObject obj,
+                 Handle<N> name,
+                 JSObject* getter,
```
bzbarsky, line 247: getter and setter are gc hazards here: they're raw pointers held across possible GC in JS_WrapValue/JS_ValueToId.

They should be passed in as handles, not raw pointers.

bzbarsky:
Apart from that rooting issue, looks good.

ehsan:
Thanks, will fix the rooting issues and land.

ehsan:
Merged manually: https://github.com/mozilla/spidernode/compare/90845bad3a066bfad8a872c9865585c7022aac93...ca1b562b80a78827af9430e4000a9384508c5ff6

PR #170, opened 2016-07-21
spidershim: fix calling Value::ToInteger on undefined

tbsaunde:


ehsan:
The changes look good, but please add some tests!

PR #171, opened 2016-07-21
Update spidermonkey

brendandahl:
First commit is updating/removing patches.
Second is the result of the update script.
Third is fixes for jsapi.

I imagine we'll want to do some more cleanup when jsruntime/context are fully merged into one.


ehsan:
Thanks, this looks great!

I have one question: What's the reason for adding the `ReportException()` helper?  It seems like that helper is changing the logic a bit, but I'm not sure if I understand why...

brendandahl:
> I have one question: What's the reason for adding the ReportException() helper? It seems like that helper is changing the logic a bit, but I'm not sure if I understand why..

JS_ReportPendingException is not in the jsapi anymore.

ehsan:
Oh, makes sense!

So there are failures on Travis.  [This](https://travis-ci.org/mozilla/spidernode/jobs/146510798) is fixed by d03c651 which I have pushed to master.  Looks like you need to upgrade to a newer version of clang for the rest though.

ehsan:
Note that the error message saying clang 3.4 is required is a [lie](https://hg.mozilla.org/integration/mozilla-inbound/rev/6be23c9426b7).  The minimum required is 3.6.

ehsan:
As noted on IRC, f6967832d4b542a6b771eee0981fd1ae1a26d6b3 upgraded clang on OSX to 3.8.

PR #182, opened 2016-08-11
add flag to set engine to spidermonkey in vcbuild

mykmelez:
@ehsan I'm not sure if the same configure flags apply to both ChakraCore and SpiderMonkey. Presumably _--without-v8-platform_ and _--without-bundled-v8_ do, at least.


mykmelez:
Note: to invoke vcbuild from MozillaBuild, use cmd, i.e.:

`cmd //c vcbuild nosign spidermonkey`

Node builds in the source directory on Windows, it doesn't build in an _out_ dir, so the binary ends up in Release/node or Debug/node.

Node docs on building on Windows:

https://github.com/nodejs/node/blob/master/BUILDING.md#windows

PR #188, opened 2016-08-13
Fix README typo.

DylanPiercey:
### Description of change

`More specifically [these tests](https://github.com/mozilla/spidernode/blob/master/deps/spidershim/test) currently passes` -> More specifically [these tests](https://github.com/mozilla/spidernode/blob/master/deps/spidershim/test) are currently passing


mykmelez:
Good catch, thanks, @DylanPiercey!

PR #193, opened 2016-09-02
also set OTHER_LDFLAGS for Xcode on Mac; link debug build against debug binaries (fixes #197)

### LINKED ISSUE #197: linker fails to link debug build due to missing symbols (including JS_SetGCZeal)
After configuring with `--debug` and `--with-external-spidermonkey-debug`:

``` bash
> ./configure --engine=spidermonkey --debug --with-external-spidermonkey-release=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim --with-external-spidermonkey-debug=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim-debug --debug
```

The linker fails on errors when linking the debug build, including (but not limited to) _JS_SetGCZeal_ being undefined:

> make
> /Applications/Xcode.app/Contents/Developer/usr/bin/make -C out BUILDTYPE=Release V=1
> make[1]: Nothing to be done for `all'.
> ln -fs out/Release/node node
> /Applications/Xcode.app/Contents/Developer/usr/bin/make -C out BUILDTYPE=Debug V=1
>   c++ -Wl,-no_pie -Wl,-search_paths_first /Users/myk/Projects/spidernode/out/Debug/spidermonkey/Debug/icudata.o -lnspr4 -mmacosx-version-min=10.9 -arch x86_64 -L/Users/myk/Projects/spidernode/out/Debug -L/Users/myk/Projects/spidernode/out/Debug -L/Users/myk/Projects/spidernode/out/Debug/spidermonkey/Debug  -o "/Users/myk/Projects/spidernode/out/Debug/arraybuffer" /Users/myk/Projects/spidernode/out/Debug/obj.target/arraybuffer/deps/spidershim/test/arraybuffer.o /Users/myk/Projects/spidernode/out/Debug/libspidershim.a /Users/myk/Projects/spidernode/out/Debug/libgtest.a /Users/myk/Projects/spidernode/out/Debug/libzlib.a -ljs_static -lz -lmozglue -lspidershim
> ld: warning: could not create compact unwind for _ffi_call_unix64: does not use RBP or RSP based frame
> Undefined symbols for architecture x86_64:
>   "JS_SetGCZeal(JSContext_, unsigned char, unsigned int)", referenced from:
>       v8::Context::CreateGlobal(JSContext_, v8::Isolate_, v8::Localv8::ObjectTemplate) in libspidershim.a(v8context.o)
>   "JSAutoCompartment::JSAutoCompartment(JSContext_, JSObject_, mozilla::detail::GuardObjectNotifier&&)", referenced from:
>       v8::Object::Set(v8::Localv8::Context, v8::Localv8::Value, v8::Localv8::Value, v8::PropertyAttribute, bool) in libspidershim.a(v8object.o)
>       v8::Object::Get(v8::Localv8::Context, v8::Localv8::Value) in libspidershim.a(v8object.o)
>       v8::Context::Impl::RunMicrotasks() in libspidershim.a(v8context.o)
>       v8::Isolate::GetHiddenGlobal() in libspidershim.a(v8isolate.o)
>       v8::ObjectTemplate::NewInstance(v8::Localv8::Object, v8::ObjectTemplate::ObjectType) in libspidershim.a(v8objecttemplate.o)
>       void mozilla::Maybe<JSAutoCompartment>::emplace<JSContext_&, JS::Rooted<JSObject*>&>(JSContext_&&&, JS::Rooted<JSObject_>&&&) in libspidershim.a(v8objecttemplate.o)
>   "JSAutoCompartment::JSAutoCompartment(JSContext_, JSObject_, mozilla::detail::GuardObjectNotifier&&)", referenced from:
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8typedarray.o)
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8arraybuffer.o)
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8arraybufferview.o)
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8object.o)
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8script.o)
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8string.o)
>       AutoJSAPI::AutoJSAPI(JSContext_, JSObject_) in libspidershim.a(v8context.o)
>       ...
>   "JS::AutoAssertOnGC::AutoAssertOnGC()", referenced from:
>       JS::AutoCheckCannotGC::AutoCheckCannotGC() in libspidershim.a(v8string.o)
>   "JS::AutoAssertOnGC::~AutoAssertOnGC()", referenced from:
>       JS::AutoCheckCannotGC::~AutoCheckCannotGC() in libspidershim.a(v8string.o)
>   "JS::detail::AssertArgumentsAreSane(JSContext_, JS::HandleJS::Value)", referenced from:
>       JS::ToNumber(JSContext_, JS::HandleJS::Value, double_) in libspidershim.a(v8value.o)
>       JS::ToInt32(JSContext_, JS::HandleJS::Value, int_) in libspidershim.a(v8value.o)
>       JS::ToUint32(JSContext_, JS::HandleJS::Value, unsigned int_) in libspidershim.a(v8value.o)
>       JS::ToString(JSContext_, JS::HandleJS::Value) in libspidershim.a(v8value.o)
>       JS::ToObject(JSContext_, JS::HandleJS::Value) in libspidershim.a(v8value.o)
>       JS::ToString(JSContext_, JS::HandleJS::Value) in libspidershim.a(v8function.o)
>       JS::ToObject(JSContext_, JS::HandleJS::Value) in libspidershim.a(v8function.o)
>       ...
>   "js::oom::failAlways", referenced from:
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8object.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8string.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8context.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8isolate.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8objecttemplate.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8function.o)
>   "js::oom::targetThread", referenced from:
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8object.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8string.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8context.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8isolate.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8objecttemplate.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8function.o)
>   "js::oom::GetThreadType()", referenced from:
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8object.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8string.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8context.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8isolate.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8objecttemplate.o)
>       js::oom::IsThreadSimulatingOOM() in libspidershim.a(v8function.o)
>   "js::oom::maxAllocations", referenced from:
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8object.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8string.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8context.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8isolate.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8objecttemplate.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8function.o)
>   "js::oom::counter", referenced from:
>       js::oom::ShouldFailWithOOM() in libspidershim.a(v8object.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8object.o)
>       js::oom::ShouldFailWithOOM() in libspidershim.a(v8string.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8string.o)
>       js::oom::ShouldFailWithOOM() in libspidershim.a(v8context.o)
>       js::oom::IsSimulatedOOMAllocation() in libspidershim.a(v8context.o)
>       js::oom::ShouldFailWithOOM() in libspidershim.a(v8isolate.o)
>       ...
> ld: symbol(s) not found for architecture x86_64
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> make[1]: *_\* [/Users/myk/Projects/spidernode/out/Debug/arraybuffer] Error 1
> make: **\* [node_g] Error 2
> 09-06 17:37 > 

Over in https://github.com/mozilla/spidernode/pull/193#issuecomment-245130580, @ehsan thought this was unrelated to the bug in spidermonkey-external.gyp that copies release binaries to the debug directory, and he asked me to file a separate issue, hence this issue. But in https://github.com/mozilla/spidernode/pull/193#issuecomment-245135188, I explained why this is indeed caused by that bug, which #193 fixes.

mykmelez:
This is needed to link against an external SpiderMonkey build from mozilla-central/gecko-dev. (Unsure why it isn't needed when the "external" build is the internal one like in the Travis tests.)

### DIFF
--- a/configure
+++ b/configure
@@ -967,6 +967,8 @@ def configure_spidermonkey(o):
       raise Exception('--with-external-spidermonkey-debug passed without specifying --with-external-spidermonkey-release')
   if sm_debug != "":
     o['variables']['external_spidermonkey_debug'] = options.external_spidermonkey_debug
+  else:
+    o['variables']['external_spidermonkey_debug'] = options.external_spidermonkey_release
   o['variables']['external_spidermonkey_release'] = options.external_spidermonkey_release
 
 
--- a/deps/spidershim/spidermonkey-external.gyp
+++ b/deps/spidershim/spidermonkey-external.gyp
@@ -112,6 +112,7 @@
               ['external_spidermonkey_release_has_nspr == 1', {
                 # Normally we'd use libraries here, but gyp doesn't allow us.
                 'ldflags': [ '-lnspr4' ],
+                'xcode_settings': {'OTHER_LDFLAGS': ['-lnspr4']},
               }],
             ],
           },
@@ -125,6 +126,7 @@
               ['external_spidermonkey_debug_has_nspr == 1', {
                 # Normally we'd use libraries here, but gyp doesn't allow us.
                 'ldflags': [ '-lnspr4' ],
+                'xcode_settings': {'OTHER_LDFLAGS': ['-lnspr4']},
               }],
             ],
           },
--- a/deps/spidershim/spidermonkey-external.gyp
+++ b/deps/spidershim/spidermonkey-external.gyp
@@ -1,9 +1,9 @@
 {
   'variables': {
-    'external_spidermonkey_debug%': '<(external_spidermonkey_release)',
+    'external_spidermonkey_debug': '<(external_spidermonkey_release)',
     'spidermonkey_binaries_debug': [
-      '<(external_spidermonkey_debug%)/js/src/<(STATIC_LIB_PREFIX)js_static<(STATIC_LIB_SUFFIX)',
-      '<(external_spidermonkey_debug%)/config/external/icu/data/icudata.o',
+      '<(external_spidermonkey_debug)/js/src/<(STATIC_LIB_PREFIX)js_static<(STATIC_LIB_SUFFIX)',
+      '<(external_spidermonkey_debug)/config/external/icu/data/icudata.o',
     ],
     'spidermonkey_binaries_release': [
       '<(external_spidermonkey_release)/js/src/<(STATIC_LIB_PREFIX)js_static<(STATIC_LIB_SUFFIX)',
@@ -12,7 +12,7 @@
     'conditions': [
       ['external_spidermonkey_debug_has_nss == 1', {
         'spidermonkey_binaries_debug': [
-          '<(external_spidermonkey_debug%)/dist/lib/<(SHARED_LIB_PREFIX)nss3<(SHARED_LIB_SUFFIX)',
+          '<(external_spidermonkey_debug)/dist/lib/<(SHARED_LIB_PREFIX)nss3<(SHARED_LIB_SUFFIX)',
         ],
       }],
       ['external_spidermonkey_release_has_nss == 1', {
@@ -22,7 +22,7 @@
       }],
       ['OS == "linux"', {
         'spidermonkey_binaries_debug+': [
-          '<(external_spidermonkey_debug%)/mozglue/build/<(STATIC_LIB_PREFIX)mozglue<(STATIC_LIB_SUFFIX)',
+          '<(external_spidermonkey_debug)/mozglue/build/<(STATIC_LIB_PREFIX)mozglue<(STATIC_LIB_SUFFIX)',
         ],
         'spidermonkey_binaries_release+': [
           '<(external_spidermonkey_release)/mozglue/build/<(STATIC_LIB_PREFIX)mozglue<(STATIC_LIB_SUFFIX)',
@@ -30,7 +30,7 @@
       }],
       ['OS == "linux" and external_spidermonkey_debug_has_nspr == 1', {
         'spidermonkey_binaries_debug+': [
-          '<(external_spidermonkey_debug%)/config/external/nspr/pr/<(SHARED_LIB_PREFIX)nspr4<(SHARED_LIB_SUFFIX)',
+          '<(external_spidermonkey_debug)/config/external/nspr/pr/<(SHARED_LIB_PREFIX)nspr4<(SHARED_LIB_SUFFIX)',
         ],
       }],
       ['OS == "linux" and external_spidermonkey_release_has_nspr == 1', {
@@ -40,7 +40,7 @@
       }],
       ['OS == "mac"', {
         'spidermonkey_binaries_debug': [
-          '<(external_spidermonkey_debug%)/dist/bin/<(SHARED_LIB_PREFIX)mozglue<(SHARED_LIB_SUFFIX)',
+          '<(external_spidermonkey_debug)/dist/bin/<(SHARED_LIB_PREFIX)mozglue<(SHARED_LIB_SUFFIX)',
         ],
         'spidermonkey_binaries_release': [
           '<(external_spidermonkey_release)/dist/bin/<(SHARED_LIB_PREFIX)mozglue<(SHARED_LIB_SUFFIX)',
@@ -48,7 +48,7 @@
       }],
       ['OS == "mac" and external_spidermonkey_debug_has_nspr == 1', {
         'spidermonkey_binaries_debug': [
-          '<(external_spidermonkey_debug%)/dist/lib/<(SHARED_LIB_PREFIX)nspr4<(SHARED_LIB_SUFFIX)',
+          '<(external_spidermonkey_debug)/dist/lib/<(SHARED_LIB_PREFIX)nspr4<(SHARED_LIB_SUFFIX)',
         ],
       }],
       ['OS == "mac" and external_spidermonkey_release_has_nspr == 1', {
--- a/deps/spidershim/spidermonkey-external.gyp
+++ b/deps/spidershim/spidermonkey-external.gyp
@@ -1,6 +1,5 @@
 {
   'variables': {
-    'external_spidermonkey_debug': '<(external_spidermonkey_release)',
     'spidermonkey_binaries_debug': [
       '<(external_spidermonkey_debug)/js/src/<(STATIC_LIB_PREFIX)js_static<(STATIC_LIB_SUFFIX)',
       '<(external_spidermonkey_debug)/config/external/icu/data/icudata.o',


mykmelez:
> This is needed to link against an external SpiderMonkey build from mozilla-central/gecko-dev. (Unsure why it isn't needed when the "external" build is the internal one like in the Travis tests.)

Ah, perhaps because the Travis tests don't build on Mac.

mykmelez:
@ehsan: The first commit on this branch, 5ddd43e, fixed building a release build with an external SpiderMonkey. But when I tried to build a debug build, I still saw linking errors (f.e. _JS_GetGCZeal_ wasn't found).

Eventually I realized that _spidermonkey.target.mk_ was copying the release libraries to the debug directories. I.e. given a configure command like this, which configures both release and debug builds:

``` bash
./configure --engine=spidermonkey --debug --with-external-spidermonkey-release=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim --with-external-spidermonkey-debug=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim-debug
```

_spidermonkey.target.mk_ contained identical prerequisites for the _$(builddir)/spidermonkey/Debug/libjs_static.a_ and _$(builddir)/spidermonkey/Release/libjs_static.a_ rules:

``` Makefile
$(builddir)/spidermonkey/Debug/libjs_static.a: /Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim/js/src/libjs_static.a FORCE_DO_CMD
    $(call do_cmd,copy)
…
$(builddir)/spidermonkey/Release/libjs_static.a: /Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim/js/src/libjs_static.a FORCE_DO_CMD
    $(call do_cmd,copy)
```

I'm still not sure exactly why, but this appears to be caused by the conditional declaration of _external_spidermonkey_debug%_ in _spidermonkey-external.gyp_:

```
'external_spidermonkey_debug%': '<(external_spidermonkey_release)',
```

According to [Providing Default Values for Variables (%)](https://gyp.gsrc.io/docs/InputFormatReference.md#Providing-Default-Values-for-Variables), this should set _external_spidermonkey_debug_ to the value of _external_spidermonkey_release_ (if the former isn't already defined). That's how node.gyp conditionally defines a bunch of variables if they aren't defined by config.gypi.

But in this case, it instead sets _external_spidermonkey_debug%_ (a variable whose name ends in `%`) to the value of _external_spidermonkey_release_, ignoring any existing value of _external_spidermonkey_debug_. After which the references to _external_spidermonkey_debug%_ in the file interpolate to the value of _external_spidermonkey_release_.

Perhaps variables are merged differently when they're in files that are included (like config.gypi is included into node.gyp), which happens early (per [Including Other Files](https://gyp.gsrc.io/docs/InputFormatReference.md#Including-Other-Files)) than when they're in files that are dependencies (like spidermoney-external.gyp is a dependency of spidershim.gyp), which may happen later.

In this case, GYP seems to treat the `%` as a literal character that is part of the name and apply this rule from [Merge Basics](https://gyp.gsrc.io/docs/InputFormatReference.md#Merge-Basics):
- When merging a dictionary, for each key in the source:
  - If the key does exist:
    - If the associated value is a string or integer, the destination value is replaced by the source value.

Thus the fix, per aae6947, is to set _external_spidermonkey_debug_ (no `%`) to the value of _external_spidermonkey_release_ in spidermonkey-external.gyp, which will set it to the value of _external_spidermonkey_debug_ in config.gypi if it exists, and otherwise will set it to the value of _external_spidermonkey_release_.

Then the configure command above will correctly copy debug libraries to the debug directories, while a configure command like this that only configures a release build still succeeds:

``` bash
./configure --engine=spidermonkey --with-external-spidermonkey-release=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim
```

ehsan:
> @ehsan: The first commit on this branch, 5ddd43e, fixed building a release build with an external SpiderMonkey. But when I tried to build a debug build, I still saw linking errors (f.e. JS_GetGCZeal wasn't found).

FWIW, your linker error is almost definitely unrelated to the mismatch between debug and release variables.  Please file an issue about that, and provide the exact linker output.  This is almost certainly a mismatch between SpiderMonkey being compiled without gc-zeal and SpiderShim being compiled with, but your configure line doesn't include `--enable-gczeal`, which makes that theory not make sense!

> 'external_spidermonkey_debug%': '<(external_spidermonkey_release)',
> 
> According to Providing Default Values for Variables (%), this should set external_spidermonkey_debug to the value of external_spidermonkey_release (if the former isn't already defined). That's how node.gyp conditionally defines a bunch of variables if they aren't defined by config.gypi.

So this part is wrong.  The meaning of the percent sign is exactly what you said above.  _But_, this also creates a variable named `foo%` unconditionally.  So the bug here is not the conditional setting of this variable, but in referencing `foo%` instead of `foo` everywhere else as you have noticed.  Can you please submit a fix for that?

(FWIW, the reason that we need to do this dance is that we cannot generate anything based on the build configuration at gyp time, so we have to unconditionally generate the Debug versions of SpiderMonkey directories in builds without `--debug`, just so that gyp can generate the same kind of build backend in both cases.)

mykmelez:
> FWIW, your linker error is almost definitely unrelated to the mismatch between debug and release variables. Please file an issue about that, and provide the exact linker output. This is almost certainly a mismatch between SpiderMonkey being compiled without gc-zeal and SpiderShim being compiled with, but your configure line doesn't include --enable-gczeal, which makes that theory not make sense!

According to https://dxr.mozilla.org/mozilla-central/rev/8c9c4e816e86f903c1d820f3f29715dc070a5a4a/js/src/old-configure.in#1829-1838, JS_GC_ZEAL is automagically enabled if MOZ_DEBUG, which is presumably why the linker tries to link it for a debug build (and not, for a release build).

In my case, the SpiderMonkey debug build is compiled with gc-zeal (automagically, because MOZ_DEBUG is defined), and the SpiderShim debug build is also compiled with gc-zeal (again, automagically, for the same reason).

The problem is that the SpiderShim debug build is being linked against the SpiderMonkey release build, because the GYP configuration causes the SpiderMonkey release build to get copied into both _spidermonkey/release/_ and _spidermonkey/debug/_.

mykmelez:
> So this part is wrong. The meaning of the percent sign is exactly what you said above. But, this also creates a variable named foo% unconditionally. So the bug here is not the conditional setting of this variable, but in referencing foo% instead of foo everywhere else as you have noticed. Can you please submit a fix for that?

That's what I thought it would do! But it doesn't actually work like that. If I define _external_spidermonkey_debug%_ in the top-level _variables_ dictionary of spidermonkey-external.gyp and then reference _external_spidermonkey_debug_ everywhere else in that file, then this configuration command:

``` bash
> ./configure --engine=spidermonkey --with-external-spidermonkey-release=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim
```

Results in this error:

> gyp: Undefined variable external_spidermonkey_debug in /Users/myk/Projects/spidernode/deps/spidershim/spidermonkey-external.gyp while loading dependencies of /Users/myk/Projects/spidernode/deps/spidershim/spidershim.gyp while loading dependencies of /Users/myk/Projects/spidernode/node.gyp while trying to load /Users/myk/Projects/spidernode/node.gyp

mykmelez:
> FWIW, your linker error is almost definitely unrelated to the mismatch between debug and release variables. Please file an issue about that, and provide the exact linker output.

Ok, I filed #197. However, for the reasons I detailed above, this is almost definitely related to the mismatch between debug and release variables, and it is almost certainly not a mismatch between SpiderMonkey being compiled without gc-zeal and SpiderShim being compiled with.

Otherwise, aae6947 wouldn't fix the linker errors! (I also confirmed that the debug build of libjs_static.a defines _JS_SetGCZeal_ before I pushed that commit.)

ehsan:
> So this part is wrong. The meaning of the percent sign is exactly what you said above. But, this also creates a variable named foo% unconditionally. So the bug here is not the conditional setting of this variable, but in referencing foo% instead of foo everywhere else as you have noticed. Can you please submit a fix for that?
> 
> That's what I thought it would do! But it doesn't actually work like that. If I define external_spidermonkey_debug% in the top-level variables dictionary of spidermonkey-external.gyp and then reference external_spidermonkey_debug everywhere else in that file, then this configuration command:
> 
> > ./configure --engine=spidermonkey --with-external-spidermonkey-release=/Users/myk/Projects/gecko-dev/obj-spidermonkey-for-spidershim
> 
> Results in this error:
> 
> ```
> gyp: Undefined variable external_spidermonkey_debug in /Users/myk/Projects/spidernode/deps/spidershim/spidermonkey-external.gyp while loading dependencies of /Users/myk/Projects/spidernode/deps/spidershim/spidershim.gyp while loading dependencies of /Users/myk/Projects/spidernode/node.gyp while trying to load /Users/myk/Projects/spidernode/node.gyp
> ```

Hmm, OK this is actually ringing a bell!  Pretty sure this is why I added those percent signs to all of the usages of this variable.

I guess now the challenge is figuring out how we're supposed to do this properly in GYP.  :/

ehsan:
Or, modify our build system to change the meaning of `--debug` to mean "build in debug mode" and not have to generate both build backends at the same time.  But I've resisted hard so far to do that.

ehsan:
Another possible approach would be to rewrite `configure_spidermonkey()` in `configure` to make it assign `external_spidermonkey_debug` to `external_spidermonkey_release` in builds without `--debug`.  That way you won't have to do the % dance any more.

Actually thinking about this more now, that may be the best way to fix this.

mykmelez:
> Or, modify our build system to change the meaning of --debug to mean "build in debug mode" and not have to generate both build backends at the same time. But I've resisted hard so far to do that.

Note that you can already build only a debug build, you just have to `make -C out BUILDTYPE=Debug` instead of invoking Make in the top-level directory.

(I'm unsure about the implications for building on Windows, where _PRODUCT_DIR_ is the top-level dir rather than the `out/` subdir; but we can burn that bridge when we come to it.)

mykmelez:
> Another possible approach would be to rewrite configure_spidermonkey() in configure to make it assign external_spidermonkey_debug to external_spidermonkey_release in builds without --debug. That way you won't have to do the % dance any more.
> 
> Actually thinking about this more now, that may be the best way to fix this.

Ok, I'll try it.

mykmelez:
> I guess now the challenge is figuring out how we're supposed to do this properly in GYP. :/

The docs are vague and contradictory, but I suspect that the way aae6947 does it is actually the way it's supposed to be done when processing the GYP file of a dependency like spidermonkey-external.gyp, even though it's different from the behavior of GYP when processing a GYP include file like config.gypi.

Nevertheless, I just pushed commit 042b40e, which ensures that external_spidermonkey_debug is set in the configure script, so spidermonkey-external.gyp doesn't have to assign it a default value. It works as expected.

ehsan:
> I guess now the challenge is figuring out how we're supposed to do this properly in GYP. :/
> 
> The docs are vague and contradictory, but I suspect that the way aae6947 does it is actually the way it's supposed to be done when processing the GYP file of a dependency like spidermonkey-external.gyp, even though it's different from the behavior of GYP when processing a GYP include file like config.gypi.

Aha, yes!  This explains everything.  Thanks for figuring it out.  :-)

ehsan (approved):

PR #195, opened 2016-09-02
WIP: copy binaries to PRODUCT_DIR/ rather than PRODUCT_DIR/spidermonkey/CONFIGURATION/ (fixes #194)

### LINKED ISSUE #194: dyld: Library not loaded: @executable_path/libnss3.dylib Reason: image not found
@ehsan After #193, SpiderNode links against an external SpiderMonkey build from mozilla-central/gecko-dev (with patches) on Mac, but running the Node executable fails because it can't find libnss3.dylib:

``` bash
> ./configure --engine=spidermonkey --with-external-spidermonkey-release=/Users/myk/Projects/gecko-dev/obj-x86_64-apple-darwin15.6.0
…
> make -j8
…
> ./node
dyld: Library not loaded: @executable_path/libnss3.dylib
  Referenced from: /Users/myk/Projects/spidernode/./node
  Reason: image not found
Trace/BPT trap: 5
```

./node is a symlink to ./out/Release/node, whereas libnss3.dylib and the other shared libraries are in ./out/Release/spidermonkey/Release/.

mykmelez:
This change fixes #194. But I'm not sure it's the right fix. I still don't quite understand spidermonkey-external.gyp, which seems to segregate external SpiderMonkey build artifacts into configuration-specific subdirectories within _<(PRODUCT_DIR)/spidermonkey/_, i.e. _<(PRODUCT_DIR)/spidermonkey/Release/_ and _<(PRODUCT_DIR)/spidermonkey/Debug/_, even though _<(PRODUCT_DIR)_ is itself already configuration-specific (it's either _out/Release/_ or _out/Debug/_, depending on configuration).

### DIFF
--- a/deps/spidershim/spidermonkey-external.gyp
+++ b/deps/spidershim/spidermonkey-external.gyp
@@ -112,6 +112,7 @@
               ['external_spidermonkey_release_has_nspr == 1', {
                 # Normally we'd use libraries here, but gyp doesn't allow us.
                 'ldflags': [ '-lnspr4' ],
+                'xcode_settings': {'OTHER_LDFLAGS': ['-lnspr4']},
               }],
             ],
           },
@@ -125,6 +126,7 @@
               ['external_spidermonkey_debug_has_nspr == 1', {
                 # Normally we'd use libraries here, but gyp doesn't allow us.
                 'ldflags': [ '-lnspr4' ],
+                'xcode_settings': {'OTHER_LDFLAGS': ['-lnspr4']},
               }],
             ],
           },
--- a/deps/spidershim/spidermonkey-external.gyp
+++ b/deps/spidershim/spidermonkey-external.gyp
@@ -91,11 +91,11 @@
 
       'copies': [
         {
-          'destination': '<(PRODUCT_DIR)/spidermonkey/Debug',
+          'destination': '<(PRODUCT_DIR)',
           'files': [ '<@(spidermonkey_binaries_debug)' ],
         },
         {
-          'destination': '<(PRODUCT_DIR)/spidermonkey/Release',
+          'destination': '<(PRODUCT_DIR)',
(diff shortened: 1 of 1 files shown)


mykmelez:
Note that this is branched off #193, as it needs that fix in order to build.

mykmelez:
This is not quite right, as it doesn't work when building debug. I'll investigate further.

ehsan:
> But I'm not sure it's the right fix. I still don't quite understand spidermonkey-external.gyp, which seems to segregate external SpiderMonkey build artifacts into configuration-specific subdirectories within <(PRODUCT_DIR)/spidermonkey/, i.e. <(PRODUCT_DIR)/spidermonkey/Release/ and <(PRODUCT_DIR)/spidermonkey/Debug/, even though <(PRODUCT_DIR) is itself already configuration-specific (it's either out/Release/ or out/Debug/, depending on configuration).

See my comment about that in #193.

ehsan:
We can't do something like this, unfortunately.  The current setup is required in order to support `--debug` builds, and that explains why those builds are currently broken for you.  GYP doesn't really support what we would like to have here, because apparently some of the backends they support don't support the necessary features for implementing only processing parts of the gyp config depending on some conditions.

The right thing to do is to set DYLD_LIBRARY_PATH (or LD_LIBRARY_PATH on Linux) according to the Node configuration you'd like to run.  This is for example [done](https://github.com/mozilla/spidernode/commit/f8f085fc55593f1afa371acd226c29753e151b4a#diff-b1dd60f6ef2061d72bf0be883e9238f3R16) in `run-tests.py`.

mykmelez:
> The right thing to do is to set DYLD_LIBRARY_PATH (or LD_LIBRARY_PATH on Linux) according to the Node configuration you'd like to run. This is for example done in run-tests.py.

Ok, fair enough.

PR #200, opened 2016-09-09
remove validation of external SpiderMonkey build

mykmelez:
Per discussion with @ehsan in IRC, this branch removes validation of an external SpiderMonkey build, which prevents integration of SpiderNode with a build process in which SpiderMonkey isn't yet built at SpiderNode configure time (i.e. Positron).

This branch also replaces the auto-detection of system vs. SpiderMonkey NSS/NSPR with explicit _--with-system-nss_ and _--with-system-nspr_ flags (like in the Mozilla build system), since the SpiderMonkey libraries will also be unavailable at SpiderNode configure time.

I didn't make those flags configuration-specific, as that seemed like overkill. Thus, when building both Debug and Release builds of SpiderNode, the consumer is expected to ensure that the Release and Debug builds of an external SpiderMonkey are both built with the same settings. (When the consumer is the Mozilla build system, as with Positron, we can ensure that it builds only a single configuration of SpiderNode that matches the configuration of SpiderMonkey.)


ehsan:
Looks like this breaks the [EXTERNAL_SPIDERMONKEY tests](https://travis-ci.org/mozilla/spidernode/builds/158626537).

@mykmelez can you please investigate?  I think you need to pass --with-system-nspr in the build commands in `.travis.yml'.

configure
```diff
@@ -446,6 +446,18 @@ parser.add_option('--with-external-spidermonkey-release',
     default='',
     help='enable linking against an out of tree SpiderMonkey release build')
 
+parser.add_option('--with-system-nspr',
+    action='store',
+    dest='system_nspr',
+    default='',
+    help='use system NSPR when linking against an out-of-tree SpiderMonkey')
```
ehsan: The meaning of "system" isn't very clear to me.  What these flags are actually controlling is whether you need to have a copy of the NSPR/NSS shared libraries as a runtime dependency.  For example, if you have an out of tree Gecko build, you will need these dependencies.  If you have an external SpiderMonkey build, you won't.

Maybe `--external-spidermonkey-has-nspr` is a better name to use?

ehsan (changes requested):

configure
mykmelez: I agree that the names and descriptions are ambiguous. I used them for consistency with the `--with-system-nspr` and `--with-system-nss` configure flags in the Gecko build system, which are described in Gecko configure scripts as "use an NSPR that is already built and installed" and "use system installed NSS," and which presumably need to be configured consistently when a Gecko build drives a SpiderNode build (i.e. when Positron configures and builds SpiderNode as a static library and then links it into libxul).

I figured it would be more obvious how the configuration flags are passed from Gecko to SpiderNode configure scripts if they're given the same names. Nevertheless, I'm happy to change them (or clarify their descriptions).

ehsan: The Gecko flags serve a completely different purpose: they use a system library for NSPR and NSS when building Gecko instead of the in-tree ones.  However, here we need flags that control whether we add Gecko's NSPR and NSS libraries built in-tree as runtime dependencies.  They don't control whether the SpiderNode build system builds them (we never build them these days in the default config.)

mykmelez: Ah, thanks for clarifying. I'll rename them as you suggested.

mykmelez:
> Ah, thanks for clarifying. I'll rename them as you suggested.

I did this in 62c35db4c60189eed3c47e63e233097dfa6d2246.

> Looks like this breaks the EXTERNAL_SPIDERMONKEY tests.
> 
> @mykmelez can you please investigate? I think you need to pass --with-system-nspr in the build commands in `.travis.yml'.

Renaming the NSPR/NSS configuration flags also inverted their meanings, so the EXTERNAL_SPIDERMONKEY tests should now pass, because `external_spidermonkey_[debug|release]_has_[nss|nspr]` will be false by default.

ehsan:
Thanks!

PR #201, opened 2016-09-09
refactor external SpiderMonkey config to avoid copying/linking

mykmelez:
@ehsan This is an experiment to see if we can avoid copying an external SpiderMonkey's libraries and linking its `dist/include` directory. It modifies spidermonkey-external.gyp to reference those files at their original locations in the external SpiderMonkey objdir, which makes it possible to move all configuration-specific settings into the _all_dependent_settings > configurations_ section of the GYP file.

I haven't tested with Linux yet, and this doesn't yet account for using a system NSS/NSPR. But it seems like it should provide the same functionality while being a bit simpler and more explicit (and reducing some file copies).

It does make it more complicated to run a Node executable that you've built with an external SpiderMonkey, since you then have to add both dist/lib and dist/sdk/lib from the SpiderMonkey objdir to (DY)LD_LIBRARY_PATH, f.e.:

``` bash
> DYLD_LIBRARY_PATH=~/Projects/gecko-dev/obj-spidermonkey-for-spidershim/dist/lib:~/Projects/gecko-dev/obj-spidermonkey-for-spidershim/dist/sdk/lib ./node -e 'console.log("hello from " + process.jsEngine)'
hello from spidermonkey
```

But that isn't an issue when building Node as a static library to link with another executable, i.e. Positron, which is the primary use case for the external SpiderMonkey support. So it seems like a reasonable complication.

Anyway, let me know what you think. Perhaps you had a particular reason for wanting to copy/link these files instead of referencing the originals?


Ryuno-Ki:
Well, I've heard about [NSS](https://en.wikipedia.org/wiki/Network_Security_Services) before but you should have a glossary around for all those abbreviations. What is NSPR? DYLD?

I'm sure C/C++ developers are familiar with it. Other contributors may not (but it is probably too early for those to chime in).

You could either put the glossary in a GitHub wiki or in a README.md.

mykmelez:
> Well, I've heard about NSS before but you should have a glossary around for all those abbreviations. What is NSPR? DYLD?

[NSPR](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSPR) is the Netscape Portable Runtime, a platform-neutral API for system-level and libc-like functions. [DYLD](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/dyld.1.html) is the Mac OS dynamic linker.

> You could either put the glossary in a GitHub wiki or in a README.md.

Hmm, I guess I think of the internet as this glossary. Google, Bing, and DuckDuckGo all return relevant references to these projects within the top three results, and usually the topmost one.

These two (and NSS) are also not particularly relevant to new contributors, since the default behavior of SpiderNode is to use its internal copy of SpiderMonkey, which doesn't require knowing anything about NSPR/NSS and DYLD_LIBRARY_PATH. The support for building with an external SpiderMonkey is targeted to particular use cases, like Positron, whose developers are already familiar with the technologies in question.

Nevertheless, I'm not opposed to such a glossary! Especially if it would help you and other potential contributors to understand the project better. Perhaps you'd be interested in starting one? I'd put it into the wiki rather than the README, under a Glossary page.

Ryuno-Ki:
Since I'm not a contributor I cannot commit on the wiki and push my changes.

However, this would be a start for the glossary.

[Glossary.md.txt](https://github.com/mozilla/spidernode/files/464610/Glossary.md.txt)

mykmelez:
> Since I'm not a contributor I cannot commit on the wiki and push my changes.

Aww, I see. Then how about a Gist? I just created this one, which you can fork and modify (and then push changes back to the original via pull requests):

https://gist.github.com/mykmelez/2dca779f52467d8e3b2a004b11196482

Ryuno-Ki:
[Nope, not supported](https://help.github.com/articles/forking-and-cloning-gists/).

You can [see the forks](https://gist.github.com/mykmelez/2dca779f52467d8e3b2a004b11196482/forks) however …

However, I found an article about [changing access permissions](https://help.github.com/articles/changing-access-permissions-for-wikis/) on GitHub help pages. You should come to a consensus internally first before doing that.

Oh, and sorry for thread hijacking ^^°

mykmelez:
> Nope, not supported.
> 
> You can see the forks however …

Ah, right, pull requests aren't possible; but you can still fork and modify, and I can pull your changes into my Gist. So it's possible to collaborate, even though the process is more manual. I've pulled your latest changes into my Gist.

> However, I found an article about changing access permissions on GitHub help pages. You should come to a consensus internally first before doing that.

Right, that's a bigger change than we can make in this thread.

> Oh, and sorry for thread hijacking ^^°

No worries!

brendandahl:
This branch is needed over in positron's node integration. I was running into issues in NSPR code with ssl socket locks being null and the sockets ended up in some very strange states. After a lot of debugging, I noticed I couldn't step into some calls to PR_PushIOLayer and lldb was failing to dump some sslSockets complaining they were defined with different translation units. What appears to be happening is libnspr is loaded twice and some code used one lib and some used the other. I'm still not sure I totally understand this or why there's not a failure during startup, but this branch fixes all those issues.

ehsan:
@mykmelez I think this is nice, but obviously we can't merge it in its current state.  Are you planning on fixing the failing travis jobs?

mykmelez:
> @mykmelez I think this is nice, but obviously we can't merge it in its current state. Are you planning on fixing the failing travis jobs?

Yes, I'm working on it. My Linux compiles take a while, as they're in a VM. And it ran out of disk space yesterday. But I plan to have an updated branch today.

mykmelez:
@ehsan, the latest commits on this branch should fix the failures in automation.

I've tested on Mac and Linux, with and without NSPR/NSS, and the builds all succeed. Tests also pass on my Mac (they fail on my Linux VM with a GYP error, but that's true on master as well, so it's unrelated to this branch).

And running Node also succeeds on Mac, with and without NSPR/NSS, as well as on Linux without NSPR/NSS. It fails on Linux with NSPR/NSS, but that also happens to me on master, so it's unrelated to this branch. I filed #206 on it.

PR #205, opened 2016-10-01
link nspr4 via 'libraries' rather than 'ldflags'

mykmelez:
@ehsan, adding `-lnspr4` to _ldflags_ for an external SpiderMonkey with NSPR doesn't work on my Linux distribution (Ubuntu 14.04 LTS). The linker complains about not finding various PR_\* symbols. I suspect this is because adding the flag via _ldflags_ instead of _libraries_ causes it to appear too early in the linker's command line.

This branch fixes that problem. It has the side-effect of requiring debug/release SpiderMonkey builds to be configured the same way: either both of them have NSPR, or neither of them do. But I've imposed the same constraint in #200, and I think it's a reasonable one.

### DIFF
--- a/deps/spidershim/spidermonkey-external.gyp
+++ b/deps/spidershim/spidermonkey-external.gyp
@@ -124,7 +124,6 @@
             'conditions': [
               ['external_spidermonkey_debug_has_nspr == 1', {
                 # Normally we'd use libraries here, but gyp doesn't allow us.
-                'ldflags': [ '-lnspr4' ],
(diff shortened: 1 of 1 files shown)


mykmelez:
Note also #206, which prevents the builds from running.

ehsan:
This is now conflicting with ec0bee69, and it seems to be a subset...  Should we close this?  @mykmelez

mykmelez:
Yes, this is no longer necessary now that #201 has landed.

PR #208, opened 2016-10-05
refine external SpiderMonkey build configuration

mykmelez:
@ehsan This branch fixes one bug (makes --external-spidermonkey-has-nspr a boolean flag) and refines the external SpiderMonkey build configuration in a variety of ways:
1. Since SpiderNode doesn't actually link NSS, it doesn't need to determine whether or not the external SpiderMonkey build has it, so I removed the `--external-spidermonkey-has-nss` flag.
2. Now that SpiderNode assumes that the debug and release builds of external SpiderMonkey either both have NSPR or neither have it, it isn't necessary to differentiate `external_spidermonkey_debug_has_nspr` from `external_spidermonkey_release_has_nspr`, so I merged them into `external_spidermonkey_has_nspr`.
3. To reduce redundancy, increase consistency, and improve the order of linker settings, I switched the settings for NSPR on Mac from OTHER_LDFLAGS to the combination of a library dir and a library.
4. I removed redundant and unnecessarily conditional settings for mozglue.

So this branch is mostly code removal (plus a big comment that makes it look like it adds more lines than it does).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -276,6 +276,7 @@ matrix:
 before_install:
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then docker build -t spidernode -f deps/spidershim/scripts/Dockerfile deps/spidershim/scripts; fi
   - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew update; fi
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew unlink autoconf; fi
   - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew install autoconf213 yasm ccache; fi
   - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew install --with-clang llvm && export PATH="/usr/local/opt/llvm/bin:$PATH"; fi
 after_install:
--- a/configure
+++ b/configure
@@ -452,12 +452,6 @@ parser.add_option('--external-spidermonkey-has-nspr',
     default=False,
     help='the out-of-tree SpiderMonkey builds have a copy of NSPR')
 
-parser.add_option('--external-spidermonkey-has-nss',
-    action='store_true',
-    dest='external_spidermonkey_has_nss',
-    default=False,
-    help='the out-of-tree SpiderMonkey builds have a copy of NSS')
-
 parser.add_option('--engine',
     action='store',
     dest='engine',
@@ -953,7 +947,6 @@ def configure_spidermonkey(o):
     o['variables']['external_spidermonkey_debug'] = options.external_spidermonkey_release
   o['variables']['external_spidermonkey_release'] = options.external_spidermonkey_release
   o['variables']['external_spidermonkey_has_nspr'] = 1 if options.external_spidermonkey_has_nspr else 0
-  o['variables']['external_spidermonkey_has_nss'] = 1 if options.external_spidermonkey_has_nss else 0
 
 
 def configure_openssl(o):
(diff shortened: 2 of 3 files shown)


mykmelez:
The Mac jobs mostly failed because they couldn't link _autoconf213_. The only one that succeeded is the one with `osx_image: xcode7.3`, which explicitly unlinks _autoconf_ (2.69) before installing _autoconf213_.

This is presumably because Travis just [switched the default Mac image to Xcode 7.3](https://blog.travis-ci.com/2016-10-04-osx-73-default-image-live/) (although it's strange that both the new and old images included _autoconf_ 2.69 by default, yet only the newer image has this problem; perhaps file permissions changed between the two versions).

In any case, the fix is obvious: unlink _autoconf_ before installing _autoconf213_ for all Mac jobs; which I've done in 5a4aab7.

mykmelez:
@tbsaunde Perhaps you can review this?

tbsaunde:
On Mon, Oct 17, 2016 at 03:58:24PM -0700, Myk Melez wrote:

> @tbsaunde Perhaps you can review this?

sure, I can take a look tomorrow.  SOrry I didn't get to it sooner I
wasn't sure if this was staled or needed attention, so feel free to ping
me on anything else you want reviewed, and I'll take a pass over the
pull requests.

Trev

tbsaunde:
The autoconf patch seems harmless, but I don't understand why you had trouble with OSX building, but master seems to build fine.  Do you have any idea?

mykmelez:
> sure, I can take a look tomorrow. SOrry I didn't get to it sooner I wasn't sure if this was staled or needed attention, so feel free to ping me on anything else you want reviewed, and I'll take a pass over the pull requests.

Thanks, @tbsaunde! This should be up-to-date. It's a followup to #201, where I refactored the external SpiderMonkey build configuration to avoid copying/symlinking files. This branch just cleans up the configuration further, removing redundant/obsolete code.

mykmelez:
> The autoconf patch seems harmless, but I don't understand why you had trouble with OSX building, but master seems to build fine. Do you have any idea?

Master broke in Travis CI [build 510](https://travis-ci.org/mozilla/spidernode/builds/166555018) eight days ago. So @ehsan landed the same autoconf fix in https://github.com/mozilla/spidernode/commit/9dbe36683a50a74a0f3f657ce538296187933256, which fixed master in [build 511](https://travis-ci.org/mozilla/spidernode/builds/166754256).

Note: @ehsan didn't push that fix via a pull request, but I recommend doing so in the future, even if you don't require review for it. That generates the "paper trail" that enables other contributors (i.e. me and you) to track those changes (f.e. by getting email about them if we're "watching" the project).

tbsaunde:
On Mon, Oct 17, 2016 at 03:58:24PM -0700, Myk Melez wrote:

> @tbsaunde Perhaps you can review this?

ok, this looks good, pushed as 09d7bac0620082946dad7bb7e6c91621708ef553

My one comment would be that I don't think we need to justify linking
against libraries in dist/ because that seems like a more stable /
better place to use than a random directory in the obj dir.

tbsaunde:
On Tue, Oct 18, 2016 at 11:09:02AM -0700, Myk Melez wrote:

> ```
> The autoconf patch seems harmless, but I don't understand why you had
> trouble with OSX building, but master seems to build fine. Do you have any
> idea?
> ```
> 
> Master broke in Travis CI build 510 eight days ago. So @ehsan landed the same
> autoconf fix in 9dbe366, which fixed master in build 511.

duh, I must have read the commit message for that 20 times and never
thought about it :(

PR #212, opened 2016-11-17
make Object::Clone() more or less work for arrays

tbsaunde:
@ehsan or @bzbarsky can you take a quick look at the sanity of the property coppying stuff here?

### DIFF
--- a/deps/spidershim/src/v8object.cc
+++ b/deps/spidershim/src/v8object.cc
@@ -454,6 +454,38 @@ Local<Object> Object::Clone() {
   JSContext* cx = JSContextFromIsolate(isolate);
   AutoJSAPI jsAPI(cx, this);
   JS::RootedObject thisObj(cx, GetObject(this));
+  bool isArray;
+  if (!JS_IsArrayObject(cx, thisObj, &isArray)) {
+    return Local<Object>();
+  }
+
+  if (isArray) {
+    uint32_t length;
+    if (!JS_GetArrayLength(cx, thisObj, &length)) {
+      return Local<Object>();
+    }
+
+    JS::RootedObject copy(cx, JS_NewArrayObject(cx, length));
+    if (!copy) {
+      return Local<Object>();
+    }
+
+    JS::RootedValue val(cx);
+    for (uint32_t i = 0; i < length; i++) {
+      if (!JS_GetElement(cx, thisObj, i, &val)) {
+        return Local<Object>();
+      }
+
+      if (!JS_DefineElement(cx, copy, i, val, JSPROP_ENUMERATE)) {
+        return Local<Object>();
+      }
+    }
+
+    JS::Value copyVal;
+    copyVal.setObject(*copy);
+    return internal::Local<Object>::New(isolate, copyVal);
+  }
+
   JS::RootedObject prototype(cx);
   JS::Value cloneVal;
   if (JS_GetPrototype(cx, thisObj, &prototype)) {
--- a/deps/spidershim/test/value.cc
+++ b/deps/spidershim/test/value.cc
@@ -1374,9 +1374,24 @@ TEST(SpiderShim, Array) {
   EXPECT_EQ(42, Integer::Cast(*val.ToLocalChecked())->Value());
   EXPECT_EQ(15u, array->Length());
 
+  Local<Object> copy = array->Clone();
+  Local<Array> arrayClone = Local<Array>::Cast(copy);
+  for (int i = 0; i < 10; ++i) {
+    MaybeLocal<Value> val = array->Get(context, i);
+    EXPECT_EQ(i * i, Integer::Cast(*val.ToLocalChecked())->Value());
+  }
+  val = arrayClone->Get(context, 14);
+  EXPECT_EQ(42, Integer::Cast(*val.ToLocalChecked())->Value());
+  EXPECT_EQ(15u, arrayClone->Length());
+
   Local<String> str = array->ToString();
   String::Utf8Value utf8(str);
   EXPECT_STREQ("0,1,4,9,16,25,36,49,64,81,,,,,42", *utf8);
+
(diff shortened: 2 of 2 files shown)


deps/spidershim/src/v8object.cc
```diff
@@ -454,6 +454,38 @@ Local<Object> Object::Clone() {
...
+    }
+
+    JS::RootedObject copy(cx, JS_NewArrayObject(cx, length));
+    if (!copy) {
+      return Local<Object>();
+    }
+
+    JS::RootedValue val(cx);
+    for (uint32_t i = 0; i < length; i++) {
+      if (!JS_GetElement(cx, thisObj, i, &val)) {
```
bzbarsky, line 475: So... this won't do the "right thing" for accessor properties.  For that matter, it won't copy non-integer-named properties.  Or preserve holes.  I'm not sure what semantics you want here, though.

deps/spidershim/src/v8object.cc
```diff
@@ -454,6 +454,38 @@ Local<Object> Object::Clone() {
...
+        return Local<Object>();
+      }
+
+      if (!JS_SetElement(cx, copy, i, val)) {
```
bzbarsky: You want JS_DefineElement.  And again, not clear what you want in terms of whether property enumerability, configurability, etc is preserved.

tbsaunde:
On Thu, Nov 17, 2016 at 01:11:17PM -0800, Boris Zbarsky wrote:

> @bzbarsky commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In deps/spidershim/src/v8object.cc:
> 
> > -  }
> >   +
> > -  if (isArray) {
> > -    uint32_t length;
> > -    if (!JS_GetArrayLength(cx, thisObj, &length)) {
> > -      return Local<Object>();
> > -    }
> >   +
> > -    JS::RootedObject copy(cx, JS_NewArrayObject(cx, length));
> > -    if (!copy) {
> > -      return Local<Object>();
> > -    }
> >   +
> > -    JS::RootedValue val(cx);
> > -    for (uint32_t i = 0; i < length; i++) {
> > -      if (!JS_GetElement(cx, thisObj, i, &val)) {
> 
> So... this won't do the "right thing" for accessor properties. For that matter,
> it won't copy non-integer-named properties. Or preserve holes. I'm not sure
> what semantics you want here, though.

So,  the one consumer of Clone() looks like this.

array = val.Clone();
for (i = 0; i < array->Length(); i++)
  array[i] = array[i]->ToString();

it then proceeds to copy the strings out of js strings and throw the
array away so I don't understand why it makes a clone at all, but
anyway.

It seems like non integer name properties don't matter, however holes do
matter.

for accessor properties I guess JS_GetElement calls the getter and
returns its return value?  That's probably fine here though I suspect
that the correct thing to do would be make a shallow copy of the
descriptor and then the ToString call in this specific case would deal
with the rest.

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In deps/spidershim/src/v8object.cc:
> 
> > -    if (!JS_GetArrayLength(cx, thisObj, &length)) {
> > -      return Local<Object>();
> > -    }
> >   +
> > -    JS::RootedObject copy(cx, JS_NewArrayObject(cx, length));
> > -    if (!copy) {
> > -      return Local<Object>();
> > -    }
> >   +
> > -    JS::RootedValue val(cx);
> > -    for (uint32_t i = 0; i < length; i++) {
> > -      if (!JS_GetElement(cx, thisObj, i, &val)) {
> > -        return Local<Object>();
> > -      }
> >   +
> > -      if (!JS_SetElement(cx, copy, i, val)) {
> 
> You want JS_DefineElement. And again, not clear what you want in terms of
> whether property enumerability, configurability, etc is preserved.

ah ok.

Looking at how V8 does this stuff it looks like an object has an array
of Object\* for properties and another for elements.  When they clone
they just copy those pointers, so I believe that they are preserved,
though of course if we only care about the one user it would seem that
doesn't really matter.

bzbarsky:
> It seems like non integer name properties don't matter, however holes do matter.

How do they matter?  It's just going 0 to length, so treating holes as undefined-valued properties, not as holes...

> for accessor properties I guess JS_GetElement calls the getter and returns its return value?

That's correct.

tbsaunde:
On Thu, Nov 17, 2016 at 03:08:17PM -0800, Boris Zbarsky wrote:

> ```
> It seems like non integer name properties don't matter, however holes do
> matter.
> ```
> 
> How do they matter? It's just going 0 to length, so treating holes as
> undefined-valued properties, not as holes...

I'm not sure I understand the difference between holes and elements where
the value is undefined.

> ```
> for accessor properties I guess JS_GetElement calls the getter and returns
> its return value?
> ```
> 
> That's correct.

ok, so in theory we should probably copy the property descriptor or
something, but for the only existing user it doesn't matter.

bzbarsky:
> I'm not sure I understand the difference between holes and elements where the value is undefined.

The difference is whether gets fall through to Array.prototype, whether getOwnPropertyDescriptor claims there is a property, whether indexOf/lastIndexOf(undefined) return -1, whether map/reduce/reduceRight/reverse/some() see the undefined value,  and probably some other ones I'm forgetting.

mykmelez:
@tbsaunde Work seems to have stalled here. Is this worth keeping open, or shall we close it as incomplete and re-open if work picks back up?

mykmelez:
I'm closing this pull request, as development has stalled on the project, so I'm going to archive it, in the interest of representing its status accurately.

PR #213, opened 2016-11-19
merge upstream from nodejs/node

mykmelez:
@tbsaunde This branch merges upstream from nodejs/node. I had to resolve a bunch of conflicts, and the build currently fails on:

```
../deps/spidershim/src/v8context.cc:35:35: error: no member named 'natives' in namespace 'spidershim'
  for (auto& native : spidershim::natives) {
                      ~~~~~~~~~~~~^
```

Which appears to be the result of upstream commit https://github.com/nodejs/node/commit/e4290dec2df252cbaaf79098784730924153b182 (or merge bustage between that change and the local changes from node-chakracore). But my fu is insufficient to determine the issue, at least right now.


tbsaunde:
On Fri, Nov 18, 2016 at 05:14:27PM -0800, Myk Melez wrote:

> @tbsaunde This branch merges upstream from nodejs/node. I had to resolve a
> bunch of conflicts, and the build currently fails on:
> 
> ../deps/spidershim/src/v8context.cc:35:35: error: no member named 'natives' in namespace 'spidershim'
>   for (auto& native : spidershim::natives) {
>                       ~~~~~~~~~~~~^
> 
> Which appears to be the result of upstream commit nodejs/node@e4290de (or merge
> bustage between that change and the local changes from node-chakracore). But my
> fu is insufficient to determine the issue, at least right now.

I think the basic problem there is that we need to update v8context.cc
to deal with the new output of the node js2c.py script.  I think that
will mean doing something similar to what that commit did to
node_javascript.cc.  We might be able to simplify a little and just use
jsapi without the v8 wrapping, but I'm not sure how much that matters.

mykmelez:
> I think the basic problem there is that we need to update v8context.cc to deal with the new output of the node js2c.py script. I think that will mean doing something similar to what that commit did to node_javascript.cc. We might be able to simplify a little and just use jsapi without the v8 wrapping, but I'm not sure how much that matters.

Indeed, 77092b4 resolves the problem by updating v8context.cc to do something similar to what node_javascript.cc does. Now on to the next error, `'unicode/uchar.h' file not found`:

>c++ '-DNODE_ENGINE="spidermonkey"' '-DNODE_ENGINE_SPIDERMONKEY' '-D_DARWIN_USE_64_BIT_INODE=1' '-DV8_TARGET_ARCH_X64' '-DENABLE_DISASSEMBLER' '-DV8_I18N_SUPPORT' '-DICU_UTIL_DATA_IMPL=ICU_UTIL_DATA_STATIC' -I../deps/v8 -I../. -I/Users/myk/Projects/spidernode/out/Release/obj/gen -I../deps/v8/include  -O3 -gdwarf-2 -fstrict-aliasing -mmacosx-version-min=10.9 -arch x86_64 -Wall -Wendif-labels -W -Wno-unused-parameter -std=gnu++0x -stdlib=libc++ -fno-rtti -fno-exceptions -fno-threadsafe-statics -fno-strict-aliasing -MMD -MF /Users/myk/Projects/spidernode/out/Release/.deps//Users/myk/Projects/spidernode/out/Release/obj.target/v8_base/deps/v8/src/char-predicates.o.d.raw   -c -o /Users/myk/Projects/spidernode/out/Release/obj.target/v8_base/deps/v8/src/char-predicates.o ../deps/v8/src/char-predicates.cc
>../deps/v8/src/char-predicates.cc:8:10: fatal error: 'unicode/uchar.h' file not found
>#include "unicode/uchar.h"

mykmelez:
Current status: it builds but doesn't run, failing with:

>fs.js:312
>Error: Offset is out of bounds
>at ReadFileContext.prototype.read (fs.js:312:3)
>at readFileAfterStat (fs.js:368:3)
>Assertion failure: !joinable(), at /Users/myk/Projects/spidernode/deps/spidershim/spidermonkey/js/src/threading/Thread.h:130
>Segmentation fault: 11

(It also still fails to build with that `'unicode/uchar.h' file not found` error on my machine. I'm currently working around that by building `--with-intl=none`.)

mykmelez:
The last three commits (160eb77, e6ed063, 1c8b796) make the straightforward changes that are required by updates to the V8 API. With those changes, I can build successfully on both Linux and Mac `--with-intl=none`.

But Node still fails to run with  "Error: Offset is out of bounds" at fs.js:312. And building with the default value for `--with-intl` still fails with "fatal error: 'unicode/uchar.h' file not found". So there's still more work to do here.

Note: there are a lot of upstream changes to V8's version of v8.h, and I applied them to SpiderShim's equivalent manually. But SpiderShim's version is based on ChakraShim's, which is very different from V8's. So in some cases I didn't apply a change because the SpiderShim version of the file didn't contain the code being changed. And it's possible that I messed up other changes in the process of manually applying them.

mykmelez:
Note: @tbsaunde and I chatted about this today, and he's going to take it over and figure out the remaining issues.

mykmelez:
@tbsaunde Is it worth leaving this open, or shall I close it, because you're working on this in a separate branch?

tbsaunde:
On Mon, Dec 19, 2016 at 02:53:39PM -0800, Myk Melez wrote:
> @tbsaunde Is it worth leaving this open, or shall I close it, because you're
> working on this in a separate branch?

I've pushed into master some pieces that are independent.  I probably
have another one or two of those, and then I have a merge.  This time at
least locally the merge seems to be fine, but I need to see what Travis
thinks.

It seems to me we might as well just leave it open as an this is a thing
being worked on maybe?  Its too bad you can't say these commits were
pushed to master to partially deal with this PR.

Ryuno-Ki:
@tbsaunde Travis thinks Merge Conflicts …

mykmelez:
Closing this because I ended up merging Node over in #351 with some additional fixes.

PR #214, opened 2016-11-23
sync upstream SpiderMonkey

mykmelez:
@tbsaunde Revision c8afe9e syncs SpiderMonkey with the latest revision on mozilla-central as of this morning (using a git-cinnabar clone of that repository), while revisions 9cdbf6d and 01df014 update SpiderShim to be compatible with upstream changes.

Revision 07c5df6 then commits the changes to js/src/configure that occur when I build SpiderNode (although I'm not sure why these changes don't happen when I build Firefox).

Finally, revision 7c9c3e6 removes all the Python bytecode (.pyc) files that the sync script imported, and revision 8fa1e3e updates the sync script to exclude those files by default, so we won't have to remove them manually the next time around.


mykmelez:
Some of the Travis builds are failing because of #216. Others complain:

> make: *** No rule to make target `../deps/spidershim/spidermonkey/build/appini_header.pyc', needed by `43a3f3d753a14c493037d4a0f08b2a2d9b4c0ee4.intermediate'.  Stop.

Which is strange, but perhaps it's to do with my deletion of all *.pyc files.

sideffect0:
Is there anything to do with CI settings, all requests are failing :/

mykmelez:
> Is there anything to do with CI settings, all requests are failing :/

I'm not sure what you're referring to, but only 21 of the 23 Travis build jobs failed.

As I mentioned earlier, some of them are failing because of #216. Others complain:

> make: *** No rule to make target `../deps/spidershim/spidermonkey/build/appini_header.pyc', needed by `43a3f3d753a14c493037d4a0f08b2a2d9b4c0ee4.intermediate'.  Stop.

mykmelez:
> make: *** No rule to make target ../deps/spidershim/spidermonkey/build/appini_header.pyc', needed by43a3f3d753a14c493037d4a0f08b2a2d9b4c0ee4.intermediate'. Stop.

This was because I forgot to remove the Python bytecode files from the list of files in spidermonkey-files.gyp. Revision 23e5083 fixes that problem.

mykmelez:
> Some of the Travis builds are failing because of #216.

This may be because the bootstrap.py script installs Rust but the Dockerfile doesn't update $PATH to include the directory containing the Rust compiler. Revision 4195bdb tries to fix that by guessing the path (based on the path on my local machine).

That revision also adds a `--verbose` option to the bootstrap.py invocation, since that invocation doesn't log where it installs Rust. But it turns out that the bootstrap.py script doesn't have the `--verbose` option (unlike `./mach bootstrap`, which does have that option, and which I was using to test locally); so revision e7f9b0f reverts that change.

We'll just have to hope that I guessed the path correctly. Otherwise, I'll have to create a Docker container locally to figure it out.

mykmelez:
Now the builds are succeeding, on both Linux and Mac (although Mac builds haven't all completed yet), but many (though not all) of the Linux builds are failing in the first test executable (exception):
>Running main() from gtest_main.cc
>[==========] Running 7 tests from 1 test case.
>[----------] Global test environment set-up.
>[----------] 7 tests from SpiderShim
>[ RUN      ] SpiderShim.ErrorConstruction
>/build/out/Release/exception failed, see the log above

mykmelez:
@tbsaunde suggested on IRC that the remaining failures are related to the old-configure.in patch, which changes the way old-configure determines the value of MOZ_GLUE_IN_PROGRAM (unsetting it unconditionally if JS_STANDALONE, which I'm guessing is defined for SpiderNode).

I would expect an update to old-configure.in to trigger recreation of old-configure, but perhaps that isn't happening. Maybe the file creation timestamps are the same, so Make doesn't think that old-configure is out-of-date?

I don't know why we even commit old-configure (and other build products of *.in files) to source control, although Mozilla has been doing that for forever (and in mozilla-central we even .gitignore the file, so you can't tell when it gets changed). Presumably there's a good reason for it.

In any case, I've recreated old-configure using the patched version of old-configure.in and committed the new version in ebb4398. If a stale old-configure was the problem with those failing Linux test runs in CI, then ebb4398 should fix it.

(SpiderNode builds and runs fine locally with that change, on both Linux and Mac. But it worked for me before the change too. I couldn't reproduce the failures in CI locally. So I don't know for sure that this is the problem.)

mykmelez:
> I don't know why we even commit old-configure (and other build products of *.in files) to source control, although Mozilla has been doing that for forever (and in mozilla-central we even .gitignore the file, so you can't tell when it gets changed). Presumably there's a good reason for it.

Erm, actually we don't. I'm not sure why I thought we did. Perhaps we used to? In any case, even if we did at one time, we don't anymore. So old-configure shouldn't be part of the sync. I've fixed that.

mykmelez:
Seven commits later, I've addressed some additional sync issues, in particular:
* don't sync js/src/configure either
* don't sync .DS_Store files if they pop up on Mac (as they did on mine)
* don't create backup (.orig) files when applying patches

Meanwhile, the last Travis CI run mostly passed, except:
* some Mac builds are still pending because of a Travis CI Mac outage earlier today
* the Linux debug asan build is failing because of memory leaks: https://travis-ci.org/mozilla/spidernode/jobs/179932449

mykmelez:
I did a clean checkout to see if there were any more files we shouldn't sync, and I found some *.o and *.so files created by psutil, so the latest commits remove them and ensure the syncing script doesn't sync them again.

tbsaunde:
On Tue, Nov 29, 2016 at 05:15:55PM -0800, Myk Melez wrote:
> I did a clean checkout to see if there were any more files we shouldn't sync,
> and I found some *.o and *.so files created by psutil, so the latest commits
> remove them and ensure the syncing script doesn't sync them again.

Thanks for taking care of that, I remember it was annoying me ages ago,
but I have no idea why we never got to fixing it.

tbsaunde:
I couldn't get a build with gcc asan to work quickly, but running valgrind on a debug build what appears to be leaking is the heldMutexStack in spidermonkey's Mutex.cpp.  Specifically it appears spidermonkey frees the stack for the main thread in JS_Shutdown(), but not the stacks for helper threads.  I'm not sure if these helper threads are actually useful to us, so not running them might be an option.  Otherwise I guess we can turn off leak checking in asan iirc there is an env var for that, or we can fix the leak in spidermonkey.

mykmelez:
> I couldn't get a build with gcc asan to work quickly, but running valgrind on a debug build what appears to be leaking is the heldMutexStack in spidermonkey's Mutex.cpp. Specifically it appears spidermonkey frees the stack for the main thread in JS_Shutdown(), but not the stacks for helper threads.

I have a Linux debug asan build now where I can reproduce the failure locally, but my stack traces are just as bad as those on Travis, so I can't yet confirm this diagnosis there. It looks like common.gypi is setting the necessary cflags (-fno-omit-frame-pointer, -fsanitize=address, -DLEAK_SANITIZER), so I'm unsure why the stack traces don't show filenames/line numbers.

I do see some additional configuration in mozconfig.asan. It's unclear if any of that is relevant.

> I'm not sure if these helper threads are actually useful to us, so not running them might be an option. Otherwise I guess we can turn off leak checking in asan iirc there is an env var for that, or we can fix the leak in spidermonkey.

It looks like mozglue/build/AsanOptions.cpp also disables asan leak detection. I guess it isn't something that Mozilla uses. So I've disabled it at runtime for the one Travis build where it's failing in d2b6657. Hopefully the Travis runs for that commit will now all pass.

It seems like we should also fix the leak upstream in SpiderMonkey, or at least report it, so the SpiderMonkey folks are aware.

@tbsaunde: can you file that as a bug and cc: me on it?

tbsaunde:
On Wed, Nov 30, 2016 at 04:49:23PM -0800, Myk Melez wrote:
>     I couldn't get a build with gcc asan to work quickly, but running valgrind
>     on a debug build what appears to be leaking is the heldMutexStack in
>     spidermonkey's Mutex.cpp. Specifically it appears spidermonkey frees the
>     stack for the main thread in JS_Shutdown(), but not the stacks for helper
>     threads.
> 
> I have a Linux debug asan build now where I can reproduce the failure locally,
> but my stack traces are just as bad as those on Travis, so I can't yet confirm
> this diagnosis there. It looks like common.gypi is setting the necessary cflags
> (-fno-omit-frame-pointer, -fsanitize=address, -DLEAK_SANITIZER), so I'm unsure
> why the stack traces don't show filenames/line numbers.

you could try running the stack through addr2line to see if that gets
you anything more maybe, or I think there is a llvm symbolizer thing you
are supposed to use?

> I do see some additional configuration in mozconfig.asan. It's unclear if any
> of that is relevant.
> 
>     I'm not sure if these helper threads are actually useful to us, so not
>     running them might be an option. Otherwise I guess we can turn off leak
>     checking in asan iirc there is an env var for that, or we can fix the leak
>     in spidermonkey.
> 
> It looks like mozglue/build/AsanOptions.cpp also disables asan leak detection.
> I guess it isn't something that Mozilla uses. So I've disabled it at runtime
> for the one Travis build where it's failing in d2b6657. Hopefully the Travis
> runs for that commit will now all pass.
> 
> It seems like we should also fix the leak upstream in SpiderMonkey, or at least
> report it, so the SpiderMonkey folks are aware.
> 
> @tbsaunde: can you file that as a bug and cc: me on it?

yeah, I took a quick look at what it would take to fix it, so I'll deal
with that.

mykmelez:
> you could try running the stack through addr2line to see if that gets you anything more maybe, or I think there is a llvm symbolizer thing you are supposed to use?

Setting LLVM_SYMBOLIZER to the location of llvm-symbolizer didn't help, and addr2line can't figure out which files/lines a given address belongs to. That suggests an issue with the build configuration, but I've tried adding the flags from mozconfig.asan (--enable-debug-symbols, --disable-install-strip, --disable-jemalloc, --disable-crashreporter, --disable-elf-hack) into build-spidermonkey.sh, and that didn't help. More research is needed.

mykmelez:
> yeah, I took a quick look at what it would take to fix it, so I'll deal with that.

Awesome, thanks!

Meanwhile, the tests finally all passed here. :-)

mykmelez:
@tbsaunde ended up doing this merge in b1c8761ebd320c0927391ef5d9cef7f3b1beb19e.

PR #218, opened 2016-12-13
Syncing SpiderMonkey from Mozilla upstream revision f8a6e712f4018e66b…

tbsaunde:


mykmelez:
One of the Travis jobs failed in the most recent Travis build, but I kicked it, and it succeeded the second time around. So Travis is green now, and this should be mergeable.

tbsaunde:
Thanks for checking

PR #347, opened 2016-12-19
sync upstream spidermonkey

mykmelez:
@tbsaunde Here's another sync with upstream SpiderMonkey. I know we synced last week, but in the interest of staying up-to-date, I plan to sync weekly (in addition to any additional ad-hoc synchronization we do in response to upstream changes).

The only bustage from this sync was an error in js/src/configure when processing the `--enable-sm-promise` flag, which was removed from SpiderMonkey in [bug 1322920](https://bugzilla.mozilla.org/show_bug.cgi?id=1322920). Fixing the bustage just required removing that flag from the *make* invocation in  build-spidermonkey.sh.


tbsaunde:
makes sense to me.  It seems better to me that we put the commits to fix things in first so we  keep history bysectable, but there's going to be other problems with that so maybe its not worth too much effort.

mykmelez:
> makes sense to me. It seems better to me that we put the commits to fix things in first so we keep history bysectable, but there's going to be other problems with that so maybe its not worth too much effort.

It's hard to land fixes first because we often won't know what needs fixing until we merge from upstream. And some of those fixes will bust SpiderMonkey until the merge, which also prevents bisectability.

I think `git bisect` can actually handle this situation pretty well, if the commits that temporarily break the application are all on a feature branch that is merged as part of a merge commit.

Nevertheless, the other thing we can do is squash the fixes into the merge commit, either before requesting a pull or when merging the pull (using GitHub's feature for squashing a pull when merging a pull).

PR #358, opened 2017-01-31
merge Node from upstream

mykmelez:
Here's the next merge from upstream Node. The only conflicts were in README.md (which always happens, since we've rewritten that file, and it changes regularly upstream) and node.gyp (due to some changes to a target that we've made conditional on `node_engine=="v8"`).

While resolving the conflicts in node.gyp, I also fixed some indentation issues that I presume we inherited from ChakraShim. Which is another reminder that it'll be helpful to get closer to upstream Node eventually.


mykmelez:
Travis was having problems with its Mac workers, so build jobs took a couple days to complete, but they're finally all done, and all jobs passed, so I'll merge this.

PR #362, opened 2017-02-13
merge Node

mykmelez:
The weekly merge. Build succeeds, and tests pass locally, although they also print this output, which I haven't seen before (and which is apparently a trace-level message in deps/zlib/trees.c):

```
bit length overflow
code 18 bits 7->6
code 6 bits 5->6

bit length overflow
code 18 bits 7->6
code 6 bits 5->6
```


mykmelez:
> they also print this output, which I haven't seen before (and which is apparently a trace-level message in deps/zlib/trees.c):

I filed that as #363.

PR #384, opened 2017-04-24
Syncing SpiderMonkey from Mozilla upstream revision d83f44d299e10fa6b0cc161ca4258c2b577f1897

mykmelez:
This is failing for me locally on Mac (and will presumably fail in Travis on Mac as well) because of [bug 1357897](https://bugzilla.mozilla.org/show_bug.cgi?id=1357897), which I'm in the process of fixing upstream.


mykmelez:
Strangely, it didn't fail, even though the version of clang on the Mac containers is 7.3.0, which should fail (and does for me locally). Go figger.

